PR #125, opened 2017-05-10
Fix module name?

gmile:


PavelVesnin:
It will fail on real working application because of Plug naming convention

PR #335, opened 2017-06-02
Metadata filtering 272

MBatrak:


lib/ehealth/web/controllers/employee_request_controller.ex
```diff
@@ -63,4 +68,36 @@ defmodule EHealth.Web.EmployeeRequestController do
...
+  defp match_employee_request(client_id, user_email, id) do
+    with %EmployeeRequest{data: data} <- API.get_by_id!(id) do
+      legal_entity_id = Map.get(data, "legal_entity_id")
+      email = get_in(data, ["party", "email"])
+      case {legal_entity_id, email} do
+        {^client_id, ^user_email} -> :ok
+        _ -> {:error, :forbidden}
+      end
+    end
+  end
```
PavelVesnin: Code between 71-102 lines should not be in controller - it's service layer.
So put it in an API of EmployeeRequest, so this code could be reused

MBatrak: Fixed

PR #346, opened 2017-06-06
Added context logic in get employees list

MBatrak:


lib/ehealth/web/controllers/employees_controller.ex
```diff
@@ -8,11 +8,15 @@ defmodule EHealth.Web.EmployeesController do
...
 
   def index(%Plug.Conn{req_headers: req_headers} = conn, params) do
-    with {:ok, %{"meta" => %{}} = response} <- PRM.get_employees(params, req_headers) do
+    client_id = get_client_id(req_headers)
+    with {:ok, %{"meta" => %{}} = response} <- PRM.get_employees(get_search_params(params, client_id), req_headers) do
       proxy(conn, response)
     end
   end
 
+  defp get_search_params(params, nil), do: params
```
PavelVesnin: client_id is always exists

PR #430, opened 2017-06-13
Added more strict rules to address validation

AlekzandrV:


lymychp:
@AlekzandrV is it in sync with apiary?

PR #431, opened 2017-06-13
Fix employee request invitation address

MBatrak:


lib/ehealth/utils/address_merger.ex
```diff
@@ -0,0 +1,137 @@
+defmodule EHealth.Utils.AddressMerger do
+  @moduledoc """
+  Address merger
+  """
+
+  alias EHealth.Dictionaries
+
+  defp config do
+    [{_pid, val}] = Registry.lookup(EHealth.Registry, :address_merger)
```
gmile: @MBatrak is there a reason to use `Registry` here? It's possible to read config during module compilation time by reading config into a module attribute

gmile: @PavelVesnin what do you think?

PavelVesnin: @gmile it's nice idea about module attributes. I think we can try it

MBatrak: Done

PR #466, opened 2017-06-19
Legal entities context changes

MBatrak:


lib/ehealth/web/controllers/legal_entity_controller.ex
```diff
@@ -22,18 +23,20 @@ defmodule EHealth.Web.LegalEntityController do
...
-    with {:ok, %{"meta" => %{}} = response} <- PRM.get_legal_entities(params, req_headers) do
-      proxy(conn, response)
+    client_id = get_client_id(req_headers)
+    params = Map.put(params, "is_active", true)
+    case Mithril.get_client_type_name(client_id) == "MIS" do
+      true ->
+        with {:ok, %{"meta" => %{}} = response} <- PRM.get_legal_entities(params, req_headers) do
+          proxy(conn, response)
+        end
+      _ -> render(conn, "index.json", legal_entities: API.get_legal_entity_list_by_id(client_id, req_headers))
```
PavelVesnin: I don't like code above because it contains too much logic, that should be in service layer.
For my opinion code in controller should be looks like this:
`case API.get_legal_entities(params, req_headers) do`
`  {:ok, %{"meta" => %{}} = response} -> proxy()`
`  {:ok, list} -> render(conn, "index.json", legal_entities: list)`
`end`

MBatrak: Done

PR #508, opened 2017-06-26
Approve declaration request

gmile:


lib/ehealth/declaration_request/api.ex
```diff
@@ -46,6 +48,17 @@ defmodule EHealth.DeclarationRequest.API do
     end
   end
 
+  def approve(id, verification_code, user_id) do
+    with declaration_request <- Repo.get!(DeclarationRequest, id) do
```
PavelVesnin: what's  the point to match with var? it's always true

PR #533, opened 2017-07-03
Update prm.erd

mmiliaieva:
Add in legal_entities table: mis_verified | enum('verified','not_verified') | not null
Change in legal_entities table status enum on ('active','closed')


specs/database_schema/prm.erd
```diff
@@ -119,8 +110,8 @@ id | uuid | not null
 name | varchar | not null
 short_name | varchar
 public_name | varchar
-type | enum('msp','mis','drug_store') | not null
-edrpou | varchar(8) | not null unique
+type | varchar | not null
+edrpou | varchar(255) | not null unique
```
PavelVesnin: why 255?

lymychp: because it has been implemented as varchar(255) on the Postgres ))
I have just synced spec with the actual implementation

PR #566, opened 2017-07-06
Updated statuses in declaration

DPashchenko:


AlekzandrV:
@DPashchenko Change status type to varchar and define Declaration state chart in Confluence page

Issue #278, opened 2017-05-29
Password management


warehousman:
As user i want to be able to reset/set new  password

lymychp:
- [ ] FE
- [ ] BE mithril

AndrewDryga:
TODOs:
- [X] API docs
- [ ] Man template design
- [ ] Man template layout
- [ ] Additional GW routes
- [X] Mithril code changes & tests
- [ ] IL code changes & tests
- [ ] Deployment

AndrewDryga:
Deployment task: https://github.com/Nebo15/ehealth.api/issues/278

AndrewDryga:
@AlekzandrV close this task after test please, but I did everything to make sure it works on dev (`curl` samples are in Slack).

AlekzandrV:
Previously created tokens remains valid after password was changed:
- [ ] Deactivate tokens when password changed

Password changed (response):
```
{
    "meta": {
        "url": "http://dev.ehealth.world/api/credentials_recovery_requests/1391a215-520f-4c83-b520-0b9f2b892b93/actions/reset_password",
        "type": "object",
        "request_id": "o8qffaao0jqo7mj2vm3hlo8himm25uav",
        "code": 200
    },
    "data": {
        "settings": {},
        "id": "fee23115-2e61-4b48-abad-f00cee8fecf3",
        "email": "sp.virny+101@gmail.com"
    }
}
```

Request Legal entities with "invalid" token:
```
curl -X GET \
  http://dev.ehealth.world/api/legal_entities \
  -H 'authorization: Bearer SGFFcERQYlR1YnoxVzRGbTFmelE0dz09' \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json' \
  -H 'postman-token: 7f4bb4d2-a134-87ab-e325-164307f6ccb9'
```

Expected result:
401

Actual result:
200

AndrewDryga:
We had a call with Alex and agreed that right now everything works as expected.

PR #635, opened 2017-07-15
Password management [minor]

AndrewDryga:
This commit contains code related to issue #278. In includes required
internal architecture.


test/web/controllers/user_controller_test.exs
```diff
@@ -0,0 +1,122 @@
...
+    end
+  end
+
+  setup do
+    {:ok, port, ref} = start_microservices(Microservices)
+
+    System.put_env("MAN_ENDPOINT", "http://localhost:#{port}")
+    System.put_env("OAUTH_ENDPOINT", "http://localhost:#{port}")
+    on_exit fn ->
+      System.delete_env("MAN_ENDPOINT")
```
gmile: Is it possible that by removing the var (instead of rolling it back to default value), depending on the execution order, a test that depends on [man](https://github.com/Nebo15/ehealth.api/blob/master/test/support/mock_server.ex#L392-L394) will fail?

(same

AndrewDryga: There is the default value in configuration, so simply removing this env is better than duplicating test value.

Also, this test is not async, which means that tests that rely on this value would fail often, which didn't happen.

PR #664, opened 2017-07-19
Cleanup declaration requests rendering & data

gmile:
Closes https://github.com/Nebo15/ehealth.api/issues/626


lib/ehealth/declaration_request/api.ex
```diff
@@ -176,9 +176,9 @@ defmodule EHealth.DeclarationRequest.API do
     |> validate_confidant_persons_tax_id()
     |> validate_confidant_person_rel_type()
     |> put_start_end_dates(global_parameters)
-    |> put_in_data("employee", employee)
-    |> put_in_data("division", division)
-    |> put_in_data("legal_entity", legal_entity)
+    |> put_in_data("employee", Create.prepare_employee_struct(employee))
```
gmile: We'll now store exactly the attributes we'll need later (e.g. during rendering a response)

PR #677, opened 2017-07-20
Science_degre.speciality shouldn't be a dic

lymychp:
just string


lib/ehealth/validators/schema_mapper.ex
```diff
@@ -98,12 +98,7 @@ defmodule EHealth.Validators.SchemaMapper do
   end
 
   def put_dictionary_value(%Dictionary{name: "SPECIALITY_TYPE", values: values}, schema, :employee_request) do
-    values = Map.keys(values)
-    schema =
-      schema
-      |> put_in(["definitions", "speciality", "properties", "speciality", "enum"], values)
-      |> put_in(["definitions", "science_degree", "properties", "speciality", "enum"], values)
-    {nil, schema}
```
PavelVesnin: @lymychp what about speciality validation for science_degree?

PR #751, opened 2017-08-02
Deduplication API process

gmile:


gmile:
Unfortunately only after opening this and reviewing the myself I now understand that `terminate_declarations`, claimed to have been implemented earlier, cannot be re-used here.

* existing implementation terminates all declarations for a given employee_id, see [this](https://github.com/edenlabllc/ops.api/blob/master/lib/ops/declaration/declarations.ex#L113-L123)
* deduplication process requires declarations to be terminated using person_id.

PR #765, opened 2017-08-04
Import PRM DB schema

gmile:
This commit adds PRM DB schema in SQL format to EHealth. It means that it is now possible to use all PRM tables when using `PRMRepo` module.

It is still required, however, to add respective Schema module before actually making queries. At the time of this commit, we have only a single schema available:

*  EHealth.PRM.GlobalParameters.Schema

Also, every time `mix test` is executed, the following line will be displayed:

```
The structure for EHealth.PRMRepo has been loaded from priv/prm_repo/structure.sql
```

Every time PRM DB changes (which is likely to happen until we fully ditch PRM API application) this file must be regenerated via the following command:

```
mix ecto.dump -r EHealth.PRMRepo
```


gmile:
Added a small fix in 55e95a1387ae1212fcc536d6e232c1307a8eb1c6

PR #789, opened 2017-08-08
Modified employee_doctors structure

AlekzandrV:


lymychp:
@AlekzandrV propose just to add `+additional_info | jsonb` to the `employee` table and deprecate employee_doctor completely.

EdwardVoloshin:
I think `employee `that it is not a high-loaded table, and therefore it will not become heavy from the new field. Logically - I agree - there is no sense in a separate table with such a structure, except for physical optimization. I agree with @AlekzandrV.

Issue #735, opened 2017-08-01
Integration with PRM database


PavelVesnin:
Migrate all PRM requests from HTTP to SQL calls using `PRMRepo`.
Some endpoints has only proxy functionality, without paging or error handling logic. In this case you should replace it from PRM.

Migration plan:
1. Rewrite HTTP calls on SQL calls with same logic (by id, lists etc.)
2. Simple refactoring for new response structures in all places, where `PRM.API` module is used
3. Replace `mock_server.ex` to `ex_machina`
4. Optimize DB calls where it needs

Step by step:
1. Choose one HTTP call in `PRM.API` module and rewrite it to SQL call (create new or update module in `prm` directory).
2. Rewrite proxy functionality if [needs](https://github.com/edenlabllc/ehealth.api/blob/master/lib/ehealth/web/controllers/division_controller.ex#L15).
3. Remove endpoint from `test/support/mock_server.ex`
4. Add `ex_machina` fixture and update tests
5. Drop outdated function from `PRM.API` module

PRM modules:
- [x] legal entity
- [x] divisions
- [x] party 
- [x] party users
- [x] employee
- [x] registry

PavelVesnin:
@gmile @Troush @AlexKovalevych welcome to discussion

gmile:
Agree with thing said above. Only want to add one thing: let's not make huge refactoring PRs. Let's do this in baby steps, one piece at a time.

lymychp:
and all the refactoring should happen only after release to prod.

gmile:
@PavelVesnin added schema in https://github.com/edenlabllc/ehealth.api/pull/765. It's now possible to use any table from PRM. You only need to add a schema file for respective table, if it doesn't exist yet

gmile:
Refactored declaration request creation to use PRM DB: https://github.com/edenlabllc/ehealth.api/pull/769

That PR introduced `legal_entity` and `medical_service_provider` schemas, now available to be used across IL.

PavelVesnin:
Refactored UkrMedRegistries: [#775](https://github.com/edenlabllc/ehealth.api/pull/775)

gmile:
Added ex_machina with support for multiple repos https://github.com/edenlabllc/ehealth.api/pull/774

gmile:
Next: going to take divisions https://github.com/edenlabllc/prm.api/blob/master/lib/prm/divisions/division.ex.

PavelVesnin:
[...]

PR #849, opened 2017-08-11
drop PRM api calls

AlexKovalevych:
https://github.com/edenlabllc/ehealth.api/issues/735
https://github.com/edenlabllc/ehealth.api/issues/856
https://github.com/edenlabllc/ehealth.api/issues/882
https://github.com/edenlabllc/ehealth.api/issues/883


lib/ehealth/declaration_request/api.ex
```diff
@@ -80,14 +79,15 @@ defmodule EHealth.DeclarationRequest.API do
...
+         %LegalEntity{} = legal_entity <- LegalEntities.get_legal_entity_by_id_with(
+                                            client_id,
+                                            :medical_service_provider
+                                          ),
+         %Division{} = division <- Divisions.get_division_by_id(attrs["division_id"])
+    do
       updates = [status: "CANCELLED", updated_at: DateTime.utc_now(), updated_by: user_id]
       global_parameters = GlobalParameters.get_values()
 
-      employee = PRMRepo.preload(employee, [:party, :doctor])
```
gmile: Where did this go?

I think it's premature to have doctor preloaded in `get_employee_by_id` call. Reason: there will definitely be more types of employees (in fact there are now, e.g. I think some form of OWNER kind of employee must exist).

By incorporating preloading `doctor` in a function with generic name, such as `get_employee_by_id`, we're making it harder for future changes, when we'd need to load employee and preload some other specific association, something other than doctor.

I'd suggest we rename `get_employee_by_id` call to something like `get_doctor_by_employee_id`, and leave `get_by_employee_id` for more generic use.

lib/ehealth/legal_entity/api.ex
```diff
@@ -17,12 +17,16 @@ defmodule EHealth.LegalEntity.API do
...
   alias EHealth.PRM.LegalEntities
   alias EHealth.PRM.LegalEntities.Schema, as: LegalEntity
+  alias Ecto.Schema.Metadata
 
   require Logger
 
   @employee_request_status "NEW"
   @employee_request_type "OWNER"
 
+  @status_closed "CLOSED"
```
gmile: :+1:

lib/ehealth/declaration_request/api.ex
AlexKovalevych: https://github.com/edenlabllc/ehealth.api/pull/849/files/90982469fa8d51574918531127f2e1ff69cb820a#diff-4b57ced23d8e589b2436ffc2985061baR55

gmile: @AlexKovalevych yeah, I figured where this code moved. I'm curious why are we making preloading `doctor` entity for all employees. Even for those who are not doctors. We might not have such non-doctor employees now, but will have them in future.

Troush: @gmile We will merge `employee_doctor` to `employee` in near future. So we can just leave it like it was before.

lib/tasks/tasks.ex
```diff
@@ -8,24 +8,25 @@ defmodule :ehealth_tasks do
...
   def migrate! do
+    prm_migrations_dir = Path.join(["priv", "prm_repo", "migrations"])
     migrations_dir = Path.join(["priv", "repo", "migrations"])
 
+    load_app()
+
+    prm_repo = EHealth.PRMRepo
+    prm_repo.start_link()
+
+    Ecto.Migrator.run(prm_repo, prm_migrations_dir, :up, all: true)
```
PavelVesnin: 👍

Troush:
🎉

PR #854, opened 2017-08-14
Update prm.erd for reimbursement main entitys

EdwardVoloshin:


specs/database_schema/prm.erd
```diff
@@ -194,6 +194,129 @@ values | jsonb | not null
...
+status | varchar | not null
+is_active | boolean | not null
+inserted_at | timestamp | not null
+inserted_by | uuid | not null
+updated_at | timestamp | not null
+updated_by | uuid | not null
+
+[medications]
+id | uuid | not null
+innm_id | uuid | not null
```
EdwardVoloshin: At [medications] column `innm_id ` - it is necessary to remove!

specs/database_schema/prm.erd
```diff
@@ -194,6 +194,129 @@ values | jsonb | not null
 labels | jsonb | not null
 is_active | boolean | not null
 
+[innms]
+id | uuid | not null
+name | varchar | not null
```
EdwardVoloshin: At `innms` column `name ` - it must be UNIQUE constraint!

specs/database_schema/prm.erd
lymychp: there should be no innm_id in the medications. Should be in the ingredients only.

PR #904, opened 2017-08-22
Change schema according to reimbursement needs

AlekzandrV:


AlexKovalevych:
Pharmacist schema implemented in separate json file

PR #917, opened 2017-08-23
Added declaration number to declaration request

AlekzandrV:


Troush:
@AlekzandrV Why we need to call it `declaration_number`? I think `declaration_id` is more explicit

PR #927, opened 2017-08-28
Refactor links generation code

gmile:


lib/ehealth/api/media_storage.ex
```diff
@@ -20,6 +20,7 @@ defmodule EHealth.API.MediaStorage do
       "bucket" => bucket,
       "resource_id" => resource_id,
       "resource_name" => resource_name,
+      "resource_content_type" => MIME.from_path(resource_name)
```
gmile: Main change in this PR.

PR #1024, opened 2017-09-12
new_substance_schema.json

EdwardVoloshin:


specs/json_schemas/new_substance_schema.json
```diff
@@ -0,0 +1,39 @@
...
+
+    "sctid_code8": {
+      "type": "string",
+      "pattern": "^[0-9]{8}$"
+    }
+  },
+  "type": "object",
+  "properties": {
+    "name": {
+      "$ref": "#/definitions/name",
```
PavelVesnin: $ref - reference is good use for large objects. In case with simple validation its preferable to use inline syntax:
```
"name": {
  "type": "string",
  "pattern": "^[a-zA-Z\\- ]*$"
}
```

PR #1119, opened 2017-09-26
Normalize ingredients model

PavelVesnin:
Separated ingredients from Medication and INNM table
Created model INNM
Refactored Medications.API -> Drugs.API


lib/ehealth/medication_dispenses/api.ex
```diff
@@ -191,9 +191,9 @@ defmodule EHealth.MedicationDispense.API do
   end
 
   defp validate_active_medication(%Medication{} = medication, %{"medication_id" => medication_id}, i) do
-    ingredient = Enum.find(medication.ingredients, &(Map.get(&1, "is_active_substance")))
-    is_valid_ingredient = Map.get(ingredient, "id") == medication_id
-    if medication.is_active && medication.type == Medication.type(:medication) && is_valid_ingredient do
+    ingredient = Enum.find(medication.ingredients, &(Map.get(&1, :is_active_substance)))
```
AlexKovalevych: what if nothing found, then you get `nil` and it will fail later on `Map.get`

lib/ehealth/prm/drugs/api.ex
```diff
@@ -0,0 +1,190 @@
...
+    :certificate,
+    :certificate_expired_at,
+  ]
+  @fields_innm_optional [:is_active]
+
+  @doc false
+  def list_medications(params) do
+    params = Map.put(params, "type", @type_medication)
+    data = %{}
+    types = %{id: Ecto.UUID, name: StringLike, form: :string, type: :string, is_active: :boolean}
```
AlexKovalevych: so you didn't create a "search" schema. I think for consistency we should either create it, or remove from all other places

lib/ehealth/prm/drugs/api.ex
```diff
@@ -0,0 +1,190 @@
...
+
+    {data, types}
+    |> cast(params, Map.keys(types))
+    |> build_search_query(INNM)
+    |> preload(:ingredients)
+    |> PRMRepo.paginate(params)
+  end
+
+  def get_innm_by_id(id), do: get_medication_entity_by_id(INNM, id)
```
AlexKovalevych: Is this used anywhere?

lib/ehealth/prm/drugs/api.ex
```diff
@@ -0,0 +1,190 @@
...
+    entity
+    |> PRMRepo.get_by!([id: id, type: entity.type()])
+    |> PRMRepo.preload(:ingredients)
+  end
+
+  def get_active_innm_by_id!(id), do: get_active_medication_entity_by_id(INNM, id)
+
+  def get_active_medication_by_id!(id), do: get_active_medication_entity_by_id(Medication, id)
+
+  defp get_active_medication_entity_by_id(entity, id) do
```
AlexKovalevych: maybe instead of a bunch of functions, its better to use something like:

```
def get_by_type_params(:medication, params), do: Medication |> where([m], ^params) |> PRMRepo.all
def get_by_type_params(:innm, params), do: INNM |> where([m], ^params) |> PRMRepo.all
```

And you call it:

```
API.get_by_type_params(:medication, [is_active: true, id: id])
```

lib/ehealth/prm/drugs/api.ex
```diff
@@ -0,0 +1,190 @@
...
+    entity
+    |> PRMRepo.get_by!([id: id, type: entity.type(), is_active: true])
+    |> PRMRepo.preload(:ingredients)
+  end
+
+  def create_innm(attrs, headers), do: create_medication_entity(INNM, attrs, headers)
```
AlexKovalevych: I prefere do it this way:

```
def create_entity(:medication, attrs, headers) do
   # validate your schema here, move common part (saving to a function)

  do_save_entity(..)
end
def create_entity(:innm, attrs, headers) do
   ...
end
```

lib/ehealth/prm/drugs/innm/ingredient.ex
```diff
@@ -0,0 +1,33 @@
...
+    field :dosage, :map
+    field :is_active_substance, :boolean, default: false
+
+    belongs_to :substance, EHealth.PRM.Drugs.Substance, type: Ecto.UUID
+    belongs_to :medication, EHealth.PRM.Drugs.Medication.Schema, type: Ecto.UUID
+
+    timestamps()
+  end
+
+  def changeset(%EHealth.PRM.Drugs.INNM.Ingredient{} = ingredient, attrs) do
```
AlexKovalevych: why wont do `def changeset(%__MODULE__{} = ingredient, attrs)` ?

lib/ehealth/prm/drugs/innm/schema.ex
```diff
@@ -0,0 +1,24 @@
+defmodule EHealth.PRM.Drugs.INNM.Schema do
+  @moduledoc false
+  use Ecto.Schema
+
+  @derive {Poison.Encoder, except: [:__meta__]}
+
+  @medication_type "INNM"
```
AlexKovalevych: not sure about this, since you know module name, you can distinguish `medication` from `innm`

lib/ehealth/web/views/medication_view.ex
```diff
@@ -11,6 +11,30 @@ defmodule EHealth.Web.MedicationView do
   end
 
   def render("medication.json", %{medication: medication}) do
-    medication
+    %{
```
AlexKovalevych: you can use `Map.take`isntead

lib/ehealth/medication_dispenses/api.ex
PavelVesnin: Ingredients with is_active_substance: true is required for Medication.

lib/ehealth/prm/drugs/api.ex
PavelVesnin: no

lib/ehealth/prm/drugs/api.ex
PavelVesnin: could be

lib/ehealth/prm/drugs/innm/ingredient.ex
PavelVesnin: 👍

lib/ehealth/web/views/medication_view.ex
PavelVesnin: 👍

lib/ehealth/prm/drugs/api.ex
PavelVesnin: It just two functions now. If amount of functions will grow, I'll rewrite this code as you propose

lib/ehealth/prm/drugs/api.ex
PavelVesnin: As I understand, in this case functions will duplicate its body, difference will be only in schema. What is a benefits?

PR #1122, opened 2017-09-26
New medications&innms model [ci skip]

EdwardVoloshin:


specs/database_schema/prm.erd
```diff
@@ -188,11 +188,11 @@ values | jsonb | not null
 labels | jsonb | not null
 is_active | boolean | not null
 
-[substances]
+[innms]
```
PavelVesnin: but why? substance are still alive

PR #1173, opened 2017-10-02
Endpoint to trigger email notification

gmile:
Will be used by OPS in order to notify authorities about
possible changes in the hash chain.


gmile:
Weird, the build didn't pass on master due to `push.sh` failure – https://travis-ci.org/edenlabllc/ehealth.api/builds/282268436. Let's see if it self-resolves.

Issue #974, opened 2017-09-01
EP: Qualify Medication request by ID


EdwardVoloshin:
Qualify Medication request by ID WS 

- [x] Create WS [spec](https://edenlab.atlassian.net/wiki/x/GIF1)
- [x] Implement
- [x] Deploy
- [x] Improve test scenarios

PR #1263, opened 2017-10-17
create medication_request qualify endpoint

AlexKovalevych:
https://github.com/edenlabllc/ehealth.api/issues/974


lib/ehealth/prm/medical_programs/schema.ex
```diff
@@ -10,6 +10,8 @@ defmodule EHealth.PRM.MedicalPrograms.Schema do
     field :inserted_by, Ecto.UUID
     field :updated_by, Ecto.UUID
 
+    has_many :program_medications, EHealth.PRM.Medications.Program.Schema
```
PavelVesnin: @AlexKovalevych you should specify foreign key as `foreign_key: : program_medications_id`, because default foreign_key name depends on last word in schema name (`Schema`). So you will get an error:
```
invalid association `program_medications` in schema EHealth.PRM.MedicalPrograms.Schema: associated schema EHealth.PRM.Medications.Program.Schema does not have field `schema_id`
```

AlexKovalevych: Actually its a warning, not an error :)

PR #1297, opened 2017-10-20
Add medication dispenses schema

AlekzandrV:


mix.exs
```diff
@@ -1,7 +1,7 @@
 defmodule EHealth.Mixfile do
   use Mix.Project
 
-  @version "0.5.130"
```
AlexKovalevych: You shouldn't do this change manually

PR #1373, opened 2017-10-30
Small changes to mrr

Troush:


lib/ehealth/medication_request_requests/api.ex
```diff
@@ -54,26 +55,45 @@ defmodule EHealth.MedicationRequestRequests do
...
+
   defp filter_by_status(query, %{"status" => status}) when is_binary(status) do
     where(query, [r], r.status == ^status)
   end
   defp filter_by_status(query, _), do: query
 
+  def show(id) do
+    mrr = get_medication_request_request(id)
+    operation = PreloadFkOperation.preload(mrr)
+    Map.merge(operation.data, %{medication_request_request: mrr})
```
PavelVesnin: why do not pipe this?
```
id
|> get_medication_request_request()
|> PreloadFkOperation.preload()
|> Map.get(:data)
|> Map.merge(%{medication_request_request: mrr})
```

Troush: `MRR` object can't be allocated in pipe

PR #1929, opened 2018-03-12
[Patch] Cabinet declaration details

ybod:


apps/ehealth/lib/ehealth/web/controllers/cabinet/declarations_controller.ex
```diff
@@ -33,4 +33,8 @@ defmodule EHealth.Web.Cabinet.DeclarationsController do
       |> send_resp(200, response)
     end
   end
+
+  def show_declaration(conn, %{"id" => id}) do
+    with {:ok, response} <- Declarations.get_declaration(id, conn.req_headers), do: proxy(conn, response)
```
AlexKovalevych: i personally don't like proxying from internal microservices, since we don't control what we send in response and any changes in ops microservice will affect response from the il

apps/ehealth/lib/ehealth/declarations/api.ex
```diff
@@ -30,6 +30,19 @@ defmodule EHealth.Declarations.API do
          do: {:ok, response}
   end
 
+  def get_declaration(id, headers) do
+    with {:ok, resp} <- OPS.get_declaration_by_id(id, headers),
```
AlexKovalevych: you can combine two lines together:

```
with {:ok, %{"data" => data}} <- ...
```

PR #1976, opened 2018-03-20
Drop Gandalf

ybod:


gmile:
Can you submit a PR to cleanup IL chart as well please?

PR #2124, opened 2018-04-10
Bugfix 2073 not all emails can be sent

progsmile:
Add Log helper
Add postmark client
Implement emails activation on employee request failure


apps/ehealth/lib/ehealth/employee_requests/employee_requests.ex
```diff
@@ -146,8 +149,11 @@ defmodule EHealth.EmployeeRequests do
             }, "$.legal_entity_id"}
          ]}
 
-      err ->
-        err
+      {:error, :mail_error, message} ->
```
progsmile: I'll remove it

PR #2150, opened 2018-04-12
[minor] Add http client abstraction on top of HTTPoison

progsmile:
Add http client abstraction on top of HTTPoison


PavelVesnin:
what benefits of this module?

PR #2197, opened 2018-04-20
Dev 2098 create regex for birth certificate

progsmile:
Add validation for birth certificate number
Update tests


apps/ehealth/lib/ehealth/persons/validator.ex
```diff
@@ -6,11 +6,13 @@ defmodule EHealth.Persons.Validator do
 
   @validation_dictionaries ["DOCUMENT_TYPE", "PHONE_TYPE", "AUTHENTICATION_METHOD", "DOCUMENT_RELATIONSHIP_TYPE"]
   @auth_method_error "Must be one and only one authentication method."
+  @birth_certificate_negitaive_regex ~r/([^A-Za-zА-яіїєґІЇЄҐ\d\#\№\–\-\—\－\_\'\,\s\/\\\=\|\!\<\;\?\%\:\]\*\+\.\√\ё])/u
 
   def validate(person) do
     dict_keys = Dictionaries.get_dictionaries_keys(@validation_dictionaries)
 
-    with %{"DOCUMENT_TYPE" => doc_types} = dict_keys,
+    with :ok <- validate_bitrh_certificate_number(person),
```
AlexKovalevych: typo in the word `bitrh`

apps/ehealth/lib/ehealth/persons/validator.ex
```diff
@@ -6,11 +6,13 @@ defmodule EHealth.Persons.Validator do
 
   @validation_dictionaries ["DOCUMENT_TYPE", "PHONE_TYPE", "AUTHENTICATION_METHOD", "DOCUMENT_RELATIONSHIP_TYPE"]
   @auth_method_error "Must be one and only one authentication method."
+  @birth_certificate_negitaive_regex ~r/([^A-Za-zА-яіїєґІЇЄҐ\d\#\№\–\-\—\－\_\'\,\s\/\\\=\|\!\<\;\?\%\:\]\*\+\.\√\ё])/u
```
AlexKovalevych: typo in the word `negitaive`

apps/ehealth/test/unit/persons/validate_person_test.exs
```diff
@@ -265,10 +265,33 @@ defmodule EHealth.Persons.ValidatorTest do
...
+      person = %{person | "birth_date" => today}
+
+      {:error, [{rules, _path}]} = Validator.validate(person)
+      assert %{description: "Must contain required item.", params: ["BIRTH_CERTIFICATE"], rule: :invalid} == rules
+    end
+
+    test "Error on invalid birth certificate number", %{person: person} do
+      person = Map.put(person, "documents", [%{"type" => "BIRTH_CERTIFICATE", "number" => "$_1232562"}])
+
+      {:error, [{rules, _path}]} = Validator.validate(person)
```
AlexKovalevych: i would use a single assert here:

```
assert {:error, [{...}]} = Validator.validate(person)
```

apps/ehealth/lib/ehealth/persons/persons.ex
```diff
@@ -66,10 +69,10 @@ defmodule EHealth.Persons do
     end
   end
 
-  defp validate_tax_id(tax_id, person_tax_id, content, signer) do
+  defp validate_tax_id(tax_id, person_tax_id, person, signer) do
     with true <- tax_id == person_tax_id,
-         true <- person_tax_id == content["tax_id"],
-         true <- content["tax_id"] == signer["drfo"] do
+         true <- person_tax_id == person["tax_id"],
```
AlexKovalevych: This looks weird, isn't it?
I mean before it was a decoded content and we compared tax_id from it with person's tax_id, now it's not possible to understand what we compare here. So either rename `person` back to `content` (i don't understand why you actually touched it) or even `signed_content` which even more accurate, if you don't like `content`, but `person` doesn't make any sense here, since it's not actually a Person.

apps/ehealth/lib/ehealth/persons/persons.ex
```diff
@@ -28,12 +30,13 @@ defmodule EHealth.Persons do
     with {:ok, %{"data" => user}} <- Mithril.get_user_by_id(user_id, headers),
          :ok <- check_user_person_id(user, id),
          %Ecto.Changeset{valid?: true, changes: changes} <- Signed.changeset(params),
-         {:ok, %{"data" => %{"content" => content, "signer" => signer}}} <-
+         {:ok, %{"data" => %{"content" => person_data, "signer" => signer}}} <-
```
AlexKovalevych: The name `person_data` is weird here (see comment below), `signed_content` or `decoded_content`, or simply `content` how it was before is a better choice.

apps/ehealth/lib/ehealth/persons/persons.ex
progsmile: `signed_content` is better in that function
It's also weird to see `JsonSchema.validate(:person, content)` - we are assuming some `content` to be valid person. I'll name it as `content`, just as it was...

apps/ehealth/test/unit/persons/validate_person_test.exs
```diff
@@ -265,29 +267,88 @@ defmodule EHealth.Persons.ValidatorTest do
...
+          # %{"type" => "OTP"},                  # only 1 of 2 at the same time
+          %{"type" => "OFFLINE"}
         ]
       },
-      "authentication_methods" => [
-        # %{"type" => "OTP"},                  # only 1 of 2 at the same time
-        %{"type" => "OFFLINE"}
-      ]
-    }
+      merge_data
```
PavelVesnin: before merge_data this function was pure. Why you can't merge your data at place where you need?
But if you need some flexible fixture, it's better create an ex_machina factory

progsmile: It's still pure)

progsmile: it produces the same output on same parameters and has no side effects

progsmile: that's much convenient to get that data what you want, instead of merging at each place

PavelVesnin: yep, my bad, it still pure, but creates not necessary merge

progsmile: yep, ex_machina factory the best choice here
I've decided to use existing code for task purposes

progsmile: we are having mpi_factory, so can move it there

PR #2206, opened 2018-04-20
[minor] implemented ExMachina params_for_string

PavelVesnin:


apps/ehealth/test/unit/persons/validate_person_test.exs
```diff
@@ -317,9 +317,10 @@ defmodule EHealth.Persons.ValidatorTest do
       ]
 
       for number <- invalid_numbers do
-        person = create_person(%{"birth_date" => @today_date, "documents" => [create_birth_certificate(number)]})
```
progsmile: there are also few places with `create_person` calling

PavelVesnin: they're for another PR

PR #2214, opened 2018-04-23
Do not pass along sensitive headers

gmile:


apps/ehealth/lib/ehealth/api/helpers/microservice_base.ex
```diff
@@ -10,6 +10,14 @@ defmodule EHealth.API.Helpers.MicroserviceBase do
 
       def process_request_options(options), do: Keyword.merge(config()[:hackney_options], options)
 
+      @filter_headers ["content-length", "Content-Length", "api-key", "authorization"]
```
gmile: After merging the HeadersProcessor into MicroserviceBase, I've added this line which was the actual aim of the PR.

PR #2219, opened 2018-04-23
Add list contract requests feature

progsmile:


apps/ehealth/lib/ehealth/contract_requests/contract_requests.ex
```diff
@@ -38,6 +43,19 @@ defmodule EHealth.ContractRequests do
     contract_number
   )a
 
+  def search(headers, client_type, params) do
+    params =
+      case client_type do
+        "MSP" -> Map.merge(params, %{"contractor_legal_entity_id" => get_client_id(headers)})
+        _ -> params
+      end
```
PavelVesnin: use [Context Plug](https://github.com/edenlabllc/ehealth.api/blob/master/apps/ehealth/lib/ehealth/plugs/client_context.ex) for adding required params.

apps/ehealth/test/web/controllers/contract_request_controller_test.exs
```diff
@@ -699,6 +701,113 @@ defmodule EHealth.Web.ContractRequestControllerTest do
...
+      assert [%{"contract_number" => @contract_number}] = response_data
+
+      %{"data" => response_data} =
+        do_get_contract_request(conn, @legal_entity_id_1, %{
+          "contractor_legal_entity_id" => @legal_entity_id_1,
+          "status" => "new"
+        })
+
+      assert [%{"contractor_legal_entity_id" => @legal_entity_id_1}, _] = response_data
+    end
```
PavelVesnin: much better, when test simple and do just one thing.
decompose it on small tests for each search query

progsmile: Do you propose to split `"finds contracts successfully"` test by assertions, so each assertion is in separate `test`?

PavelVesnin: yep

apps/ehealth/lib/ehealth/contract_requests/contract_requests.ex
```diff
@@ -38,6 +43,13 @@ defmodule EHealth.ContractRequests do
     contract_number
   )a
 
+  def search(search_params) do
+    with %Ecto.Changeset{valid?: true} = changeset <- ContractRequestsSearch.changeset(search_params),
+         %Scrivener.Page{} = paging <- search(changeset, search_params, ContractRequest) do
```
AlexKovalevych: Please use alias `Scrivener.Page` at the top of the module

apps/ehealth/lib/ehealth/contract_requests/contract_requests.ex
```diff
@@ -1,18 +1,23 @@
...
 
+  use EHealth.Search, EHealth.Repo
+
+  import Ecto.Changeset
+  import Ecto.Query
   import EHealth.Utils.Connection, only: [get_consumer_id: 1, get_client_id: 1]
+
   alias Ecto.Adapters.SQL
   alias EHealth.ContractRequests.ContractRequest
+  alias EHealth.ContractRequests.Search, as: ContractRequestsSearch
```
AlexKovalevych: I wouldn't alias it as `ContractRequestsSearch`, since you have no others searches here, the default `Search` is fine

apps/ehealth/lib/ehealth/contract_requests/search.ex
```diff
@@ -0,0 +1,26 @@
...
+    field(:contract_number, :string)
+  end
+
+  def changeset(params) do
+    %__MODULE__{}
+    |> cast(params, __MODULE__.__schema__(:fields))
+    |> uppercase_status()
+  end
+
+  defp uppercase_status(%{changes: %{status: _}} = changeset), do: update_change(changeset, :status, &String.upcase(&1))
```
AlexKovalevych: `%Ecto.Changeset{valid?: true}` is better in pattern match, since no need to uppercase status, if changeset is invalid

apps/ehealth/lib/ehealth/plugs/contract_context.ex
```diff
@@ -0,0 +1,78 @@
...
+    end
+  end
+
+  defp validate_params_list(conn, _config), do: conn
+
+  defp legal_entity_id_not_allowed?(%Plug.Conn{} = conn, legal_entity_id, personal_client_types) do
+    conn.assigns.client_type in personal_client_types and legal_entity_id != get_client_id(conn.req_headers)
+  end
+
+  defp put_context_params(
```
AlexKovalevych: Isn't this copy paste from `client_context.ex` ?

apps/ehealth/lib/ehealth/web/views/contract_request_view.ex
```diff
@@ -3,6 +3,30 @@ defmodule EHealth.Web.ContractRequestView do
 
   use EHealth.Web, :view
 
+  def render("index.json", %{contract_requests: contract_requests}) do
+    render_many(contract_requests, __MODULE__, "contract_request.json")
+  end
+
+  def render("contract_request.json", %{contract_request: contract_request}) do
```
AlexKovalevych: Consider name it `list.json` or `list_contract_request.json` to underline that it's a list item render

apps/ehealth/test/web/controllers/contract_request_controller_test.exs
```diff
@@ -699,6 +704,106 @@ defmodule EHealth.Web.ContractRequestControllerTest do
     end
   end
 
+  describe "search contract request" do
+    @nhs_signer_id UUID.generate()
+    @contract_number UUID.generate()
```
AlexKovalevych: don't understand why you need module attributes here

apps/ehealth/lib/ehealth/contract_requests/contract_requests.ex
AlexKovalevych: It's already explicit - `Search` )
Otherwise why did you name the module `Search` but not `ContractRequestsSearch` and according to your logic, you need to alias it that way each time you use it - no sense for me.

apps/ehealth/lib/ehealth/contract_requests/search.ex
progsmile: `status` is optional, so the changeset is valid, but we are getting `nil` inside `&String.upcase(&1)`

apps/ehealth/test/web/controllers/contract_request_controller_test.exs
progsmile: they are scoped by `describe` and are shared in tests below

apps/ehealth/lib/ehealth/web/views/contract_request_view.ex
progsmile: ok

PR #2256, opened 2018-05-03
Get contract details

progsmile:
Implement get contract details


apps/ehealth/lib/ehealth/contracts/contracts.ex
```diff
@@ -0,0 +1,80 @@
+defmodule EHealth.Contracts do
+  @moduledoc false
+
+  alias EHealth.Repo
+  alias EHealth.Validators.Preload
+  alias EHealth.ContractRequests.ContractRequest
+
+  @ops_api Application.get_env(:ehealth, :api_resolvers)[:ops]
+
+  @forbidden_message "You are not allowed to view this contract"
```
AlexKovalevych: I don't see any need in moving this message to module attribute. E.g. - we may need to have a different message for different process

apps/ehealth/lib/ehealth/contracts/contracts.ex
```diff
@@ -0,0 +1,80 @@
...
+
+  def get_by_id(id, params) do
+    with {:ok, %{"data" => contract}} <- @ops_api.get_contract(id, []),
+         :ok <- check_get_by_id_permitions(contract, params),
```
AlexKovalevych: i guess it should be `permissions`. But in general i would rename this function to something like `validate_contractor_legal_entity` or `validate_contractor_legal_entity_id`

apps/ehealth/lib/ehealth/contracts/contracts.ex
```diff
@@ -0,0 +1,80 @@
...
+
+  defp check_get_by_id_permitions(contract, %{"contractor_legal_entity_id" => contractor_legal_entity_id}) do
+    if contract["contractor_legal_entity_id"] == contractor_legal_entity_id,
+      do: :ok,
+      else: {:error, {:forbidden, @forbidden_message}}
+  end
+
+  defp check_get_by_id_permitions(_contract, _params), do: :ok
+
+  defp load_contract_references(contract) do
```
AlexKovalevych: Please don't generate view in this module, load only references and return them. Generate response json in the view module

apps/ehealth/lib/ehealth/contracts/contracts.ex
progsmile: ok

PR #2276, opened 2018-05-07
[Minor] Update DS logic and tests for double signed documents

ybod:


apps/ehealth/test/support/mock_server.ex
```diff
@@ -268,6 +268,8 @@ defmodule EHealth.MockServer do
       |> Map.get("signed_content")
       |> Base.decode64()
 
+    IO.inspect(data)
```
AlexKovalevych: please remove

ybod: done

Issue #2255, opened 2018-05-03
Move addresses to separate table


AlexKovalevych:
In order to optimize stats selects
Move addresses from division to separate table

svetavedmed:
some division's addresses are not added to division_addresses. 
Ex. at the moment there are 4 divisions, which addresses do not represented in division_addresses
```
division_id
091c9efa-37fd-44c9-a7d1-a5249ae7ff26
439456c7-67a2-4d8d-bf6a-ea135cfd68e6
998b224b-6ae8-443e-8f54-8bde72817174
aab1911e-4a3f-48d9-9b18-8125510eafb7
```

PR #2312, opened 2018-05-18
[minor] Duplicate division addresses to table

progsmile:
#2255


apps/ehealth/lib/ehealth/divisions/divisions.ex
```diff
@@ -61,16 +61,21 @@ defmodule EHealth.Divisions do
   end
 
   def get_by_id!(id) do
-    PRMRepo.get!(Division, id)
+    Division
+    |> query_addresses()
```
progsmile: actually such query logic supposed to be in step 2, but still it can be used

PR #2450, opened 2018-06-13
Drop simple factory #2416

magiccookie:


apps/ehealth/test/data/address_type.json
```diff
@@ -0,0 +1,14 @@
+{
```
AlexKovalevych: please use factory for address

apps/ehealth/test/support/address-factory.ex
```diff
@@ -0,0 +1,24 @@
+defmodule EHealth.AddressFactory do
+  @moduledoc false
+
+  defmacro __using__(_opts) do
+    quote do
+      def address_factory do
+        %{
+          type: "REGISTRATION",
```
AlexKovalevych: you can use string keys and avoid extra conversion

apps/ehealth/test/support/address-factory.ex
```diff
@@ -0,0 +1,24 @@
+defmodule EHealth.AddressFactory do
```
AlexKovalevych: please rename `address-factory.ex` to `address_factory.ex` and move it under `factories` directory

magiccookie: done

apps/ehealth/test/support/address-factory.ex
magiccookie: done

apps/ehealth/test/data/address_type.json
magiccookie: done

Issue #2407, opened 2018-06-08
Changes on Sign-up


svetavedmed:
EHealth:
  - add call to OTP Verification with OTP before MPI.Person creation
  - ensure that OTP valid, return validation message if needed

Mithril:
  - remove from Mithril.UserAPI.create_or_update_user_factor/2 :with_otp_validation attribute and
  - ensure that OTP not required for User create/update process with Factor

svetavedmed:
Check OTP we received is the same as was sent. Check verifications.phone_number=$phone_number and verifications.code=$code and verifications.status='new' and update status to 'verified'.
in case of error return 403 Error (`Invalid verification code`)

PR #2468, opened 2018-06-15
auth verification added (cabinet registration)

imax-iva:
https://github.com/edenlabllc/ehealth.api/issues/2407


apps/ehealth/lib/ehealth/cabinet/api.ex
```diff
@@ -270,4 +273,26 @@ defmodule EHealth.Cabinet.API do
       _error -> {:error, {:bad_gateway, "Failed to save signed content"}}
     end
   end
+
+  def verify_auth(%{"authentication_methods" => authentication_methods}, %{"otp" => code}, headers) do
+    otp_params = Enum.filter(authentication_methods, fn item -> otp_params?(item) end)
```
AlexKovalevych: why not to use `Enum.find` ?

Issue #2298, opened 2018-05-14
declaration request in status expired


svetavedmed:
As a MIS I want to see declaration request in status expired w/o data so that I could see status and ids
[#402](https://github.com/ehealth-ua/api-spec/issues/402)

svetavedmed:
request
```
curl -X GET \
  https://api.dev.asclepius.com.ua/api/declaration_requests/16093e5e-571b-4a85-be96-bd443c678c45 \
  -H 'Authorization: Bearer ME1ZMWRhWnd1NWNzUmNzbGVvYjZvQT09' \
  -H 'Cache-Control: no-cache' \
  -H 'Postman-Token: f44fdbbf-25c6-437a-a013-72d31490ec15' \
  -H 'api-key: UXhEczJacXh3Ri9SUkc4enhHTVVpQT09'
```
response
```
{"meta":{"url":"http://api-svc.il/api/declaration_requests/16093e5e-571b-4a85-be96-bd443c678c45","type":"object","request_id":"31134980-d1eb-4626-8b3c-95a884b7db1e#13110","code":500},"error":{"errors":{"detail":"Internal server error"}}}
```

svetavedmed:
- [ ] add, please, declaration_id to response

PR #2476, opened 2018-06-18
Get Declaration Request in status expired when data is null

magiccookie:
Fix error 500

#2298


apps/ehealth/lib/ehealth/declaration_requests/declaration_requests.ex
```diff
@@ -157,12 +157,16 @@ defmodule EHealth.DeclarationRequests do
     |> Repo.one()
   end
 
+  defp validate_data_legal_entity(%DeclarationRequest{data: nil} = declaration_request, _) do
+    declaration_request |> Map.put(:data, %{})
+  end
+
+  defp validate_data_legal_entity(%DeclarationRequest{data: data} = declaration_request, _) when data == %{} do
```
AlexKovalevych:
```
defp validate_data_legal_entity(%DeclarationRequest{data: data} = declaration_request, _) when data == %{} or is_nil(data) do
   Map.put(declaration_request, :data, %{})
end
```

apps/ehealth/lib/ehealth/declaration_requests/declaration_requests.ex
```diff
@@ -157,12 +157,16 @@ defmodule EHealth.DeclarationRequests do
     |> Repo.one()
   end
 
+  defp validate_data_legal_entity(%DeclarationRequest{data: nil} = declaration_request, _) do
```
AlexKovalevych: you don't need this function (see comment below)

apps/ehealth/lib/ehealth/declaration_requests/declaration_requests.ex
magiccookie: done

Issue #2432, opened 2018-06-12
error 500 on creating legal_entity with type MIS


svetavedmed:
Original task [446](https://github.com/ehealth-ua/api-spec/issues/446)

Only legal_entity with type MSP or Pharmacy could be created by MIS

svetavedmed:
"message": "Only legal_entity with type MSP or Pharmacy could be created",

PR #2488, opened 2018-06-19
[minor] Add error message on creating legal_entity with type MIS

magiccookie:
#2432


apps/ehealth/lib/ehealth/legal_entities/validator.ex
```diff
@@ -117,6 +121,14 @@ defmodule EHealth.LegalEntities.Validator do
          do: :ok
   end
 
+  def validate_type(content), do: check_type(content["type"])
```
AlexKovalevych: please use pattern matching:

```
defp validate_type(%{"type" => @msp}), do: :ok
...
```

Issue #2451, opened 2018-06-13
declaration request create online response


svetavedmed:
Responses for declaration create online and get declaration request details online should be the same

PR #2504, opened 2018-06-20
declaration request api improving

imax-iva:
https://github.com/edenlabllc/ehealth.api/issues/2451


apps/ehealth/test/web/controllers/contract_controller_test.exs
```diff
@@ -1,7 +1,7 @@
 defmodule EHealth.Web.ContractControllerTest do
   @moduledoc false
 
-  use EHealth.Web.ConnCase
+  use EHealth.Web.ConnCase, async: false
```
AlexKovalevych: not needed

Issue #2497, opened 2018-06-20
Check legal_entity on create contract request


svetavedmed:
Check if external_contractors.legal_entity_id exist and status='ACTIVE'
in case error return 422 msg "Active $external_contractors[i].legal_entity_id is not exist"

magiccookie:
This will return the error of the with the first invalid legal entity occurrence, not the list of errors.
Please, refer to @AlexKovalevych in case

svetavedmed:
```
{
                “rules”: [
                    {
                        “rule”: “invalid”,
                        “params”: [],
                        “description”: “Active $external_contractors[1].legal_entity_id does not exist”
                    }
                ],
                “entry_type”: “json_data_property”,
                “entry”: “$.contractor_employee_divisions[1].division_id”
            }
```

PR #2571, opened 2018-07-03
add check external contractors legal entity id

magiccookie:
#2497


apps/ehealth/lib/ehealth/contract_requests/contract_requests.ex
```diff
@@ -993,29 +993,66 @@ defmodule EHealth.ContractRequests do
...
-            params["contractor_divisions"],
-            contractor_division,
-            "$.external_contractors[#{i}].divisions[#{j}].id"
-          )
-        end)
-
-      case validation_result do
-        :ok ->
-          validate_external_contract(contractor, params, "$.external_contractors[#{i}].contract.expires_at")
+      legal_entity_validation = validate_external_legal_entity(contractor, i)
```
AlexKovalevych: i would do it in the way:

```
with :ok <- validate_legal_entity(contractor, i),
     :ok <- validate_divisions(params, contractor, i),
     :ok <- validate_external_contract(params, contractors, i) do
        {:cont, :ok}
else
   {:error, reason} -> {:halt, {:error, reason}}
end
```

apps/ehealth/lib/ehealth/contract_requests/contract_requests.ex
```diff
@@ -993,29 +993,66 @@ defmodule EHealth.ContractRequests do
     end
   end
 
+  defp validate_external_legal_entity(contractor, i) do
```
AlexKovalevych: you can use pattern matching here:

```
  defp validate_external_legal_entity(%{"legal_entity_id" => legal_entity_id}, i) when not is_nil(legal_entity_id) do
  defp validate_external_legal_entity(contractor, i) do
```

magiccookie: done

apps/ehealth/lib/ehealth/contract_requests/contract_requests.ex
magiccookie: done

Issue #2623, opened 2018-07-09
Declaration requests for patients with age<14


svetavedmed:
In case patient age<14 confidant_person should be mandatory in declaration request  

If person age < 14 check existence of confidant_person
 - in case error return 422 - msg "Confidant person is mandatory for children"

danakonovalenko:
@svetavedmed patient age<14 still can make declaration request without confidant person in case this patient has DS and tax_id.
For combined method (when person and declaration are in one request) tested and done

svetavedmed:
on POST http://ehealth.edenlabllc.com/api/cabinet/declaration_requests
check persons age. 
If person age < 14 check existence of confidant_person
 - in case error return 422 - msg "Confidant person is mandatory for children"

PR #2642, opened 2018-07-11
Declaration requests for patients with age < 14 

magiccookie:
#2623


apps/ehealth/lib/ehealth/persons/validator.ex
```diff
@@ -148,15 +148,37 @@ defmodule EHealth.Persons.Validator do
...
+    end
+  end
+
   defp validate_confidant_persons(person, dict_keys) when is_map(person) do
     valid_relations = ["PRIMARY", "SECONDARY"]
 
     case Map.get(person, "confidant_person") do
       nil ->
-        :ok
+        require_confidant_person_for_child(person)
```
AlexKovalevych: let's use pattern matching in validate_confidant_persons function, e.g.:

```
defmodule A do
    def validate_confidant_persons(%{"condidant_persons" => [h | t] = confidant_persons}) do
        IO.inspect(confidant_persons)
    end

    def validate_confidant_persons(_) do
        IO.inspect("NOTHING")
    end  
end
```

magiccookie: done

Issue #2768, opened 2018-07-21
Translation for scopes


danakonovalenko:
We need to make migrations on dev/demo for dictionaries.scope adding new values

```
update dictionaries
set values = values::jsonb ||'{"person:read": "Отримання даних пацієнта","reimbursement_report:read":"Отримання даних реімбурсаційного звіту",	"bl_user:write":"Додати користувача до чорного списку","bl_user:read":"Отримати дані користувачів з чорного спсику", "bl_user:deactivate":"Видалити користувача з чорного списку",	"user:block":"Заблокувати користувача","user:unblock":"Розблокувати користувача", "user:request_factor":"Запит другого фактор авторизації користувача",	"user:approve_factor":"Затвердження другого фактор авторизації користувача", "authentication_factor:write":"Зміна двофакторної аторизації",	"authentication_factor:read":"Отримання даних двофакторної аторизації",	"contract_request:create":"Створення запиту на контракт",	"contract_request:read":"Отримання деталей запиту на контракт",	"contract_request:approve":"Затвредження запиту на контракт",	"contract_request:update":"Редагування запиту на контракт",	"contract_request:sign":"Підпис запиту на контракт", "contract_request:terminate":"Відхилення запиту на контракт",	"contract:read":"Отримання деталей контракту",	"contract:write":"Оновлення контракту",	"contract:terminate":"Відхилення контракту", "capitation_report:read":"Отримання даних капітаційного звіту"}'::jsonb
where name='SCOPES'
;
```

PR #2773, opened 2018-07-23
Translation for scopes

magiccookie:
#2768


apps/ehealth/priv/repo/migrations/20180723075909_scope_tranlation.exs
```diff
@@ -0,0 +1,34 @@
+defmodule EHealth.Repo.Migrations.ScopeTranlation do
+  use Ecto.Migration
+
+  def change do
+    execute(
+      """
+      UPDATE dictionaries SET
+      values = values::jsonb ||'{
```
AlexKovalevych: this should be done via `dictionaries.json`

PR #2842, opened 2018-07-27
add sequence 

marinakr:


apps/ehealth/lib/ehealth/declaration_requests/api/creator.ex
```diff
@@ -97,7 +100,7 @@ defmodule EHealth.DeclarationRequests.API.Creator do
       {:error, %Changeset{errors: [declaration_number: {"has already been taken", []}]}} ->
         # declaration_number collision, let's regenerate it
         changeset
-        |> put_declaration_number(NumberGenerator.generate(1, 2))
+        |> put_declaration_number()
```
AlexKovalevych: this won't happen (since we use new algorithm), so the whole pattern matching block should be dropped

apps/ehealth/lib/ehealth/utils/number_generator.ex
```diff
@@ -8,9 +8,10 @@ defmodule EHealth.Utils.NumberGenerator do
   @human_readble_symbols ~w(0 A 1 E 2 H 3 K 4 M 5 P 6 T 7 X 8 9)
   # Number of symbols available, set as module attribute for better performance
   @options_count 18
-  @ver_1 "0000-"
 
-  def generate_from_sequence(1, sequence, blocks \\ 3) do
+  defp version_prefix(n), do: "000#{n}-"
+
+  def generate_from_sequence(n, sequence, blocks \\ 3) do
```
AlexKovalevych: you should fix number generation in contract_requests to use 0 version

marinakr: It use 0

PR #2915, opened 2018-08-10
Master

newlifebeat:
to fix some issues


AlexKovalevych:
We're not merging from master to develop

PR #2975, opened 2018-08-22
get grugs pagination fixed #2891

### LINKED ISSUE #2891: Reimbursement demo fix bugs BE
- [x] WS `Get Drugs ({{host}}/api/drugs)`  must return unique packages[] in response. As for now packages can be duplicated
- [x] WS `{{host}}/api/medication_request_requests/prequalify` and WS `{{host}}/api/medication_requests/{{medication_request_id}}/actions/qualify` must return response only related to innm_dosage from medication request. As for now all participants of medical program will be returned [spec](https://edenlab.atlassian.net/wiki/x/GIF1)

imax-iva:
https://github.com/edenlabllc/ehealth.api/issues/2891


apps/core/lib/core/medications/medications.ex
```diff
@@ -62,8 +62,18 @@ defmodule Core.Medications do
   end
 
   defp search_drugs(%{valid?: true, changes: attrs}, params) do
-    page_number = Map.get(params, "page", 1)
-    page_size = Map.get(params, "page_size", 50)
+    page_number =
```
AlexKovalevych: Let's set default page number in a single place:

```
page_number = 1
page_number = case Integer.parse(Map.get(params, "page", "")) do
        {int, _} -> max(int, page_number)
        :error -> page_number
      end

```

Issue #2988, opened 2018-08-27
Change code_atc to array


danakonovalenko:
On [Create Medication](https://uaehealthapi.docs.apiary.io/reference/internal.-nhs-admin/medications/create-medication) change `code_atc` from `string` to `array`.
- [x] make changes in responses:
   -  [Create Medication](
https://uaehealthapi.docs.apiary.io/reference/internal.-nhs-admin/medications/create-medication)
   - [Deactivate Medication](https://uaehealthapi.docs.apiary.io/reference/internal.-nhs-admin/medications/deactivate-medication)
   - [Get Medication By ID](
https://uaehealthapi.docs.apiary.io/reference/internal.-nhs-admin/medications/get-medication-by-id)
   - [Get Medications List](
https://uaehealthapi.docs.apiary.io/reference/internal.-nhs-admin/medications/get-medications-list)

- [x] change search by atc_code in Get Drugs (one of array medications.[code_atc]=$code_atc) (case insensitive exact search)

danakonovalenko:
- [ ] add validation that elements are sent in array are distinct
as for now it's possible to create medications with next atc_codes 
```
"code_atc": ["C01BD08","C01BD08"]
```
show 422 error: 'atc codes are duplicated'

PR #3057, opened 2018-09-03
Change code atc to array#2988

imax-iva:
https://github.com/edenlabllc/ehealth.api/issues/2988


apps/core/priv/prm_repo/migrations/20180831135805_medication_code_atc_to_array.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Core.PRMRepo.Migrations.MedicationCodeAtcToArray do
+  use Ecto.Migration
+
+  def change do
+    execute("""
```
AlexKovalevych: `@disable_ddl_transaction true`?

PR #3093, opened 2018-09-11
Implemented mithril connections

PavelVesnin:


apps/core/lib/core/legal_entities/legal_entities.ex
```diff
@@ -411,7 +398,7 @@ defmodule Core.LegalEntities do
   defp validate_msp_required(changeset), do: changeset
 
   def get_context_params(client_id, client_type, legal_entity_param_name \\ @legal_entity_param_name_default) do
-    config = Confex.fetch_env!(:core, Core.LegalEntities)
+    config = Confex.fetch_env!(:core, __MODULE__)
```
AlexKovalevych: maybe it's better to `use Confex` and use `config()` function

apps/core/lib/core/oauth/api.ex
```diff
@@ -11,6 +11,30 @@ defmodule Core.OAuth.API do
...
+
+    with {:ok, %{"data" => client}} <- @mithril_api.put_client(client_attrs, headers),
+         {:ok, %{"data" => connection}} <- create_or_update_connection(client["id"], connection_attrs, headers) do
+      {:ok, client, connection}
+    end
+  end
+
+  def create_or_update_connection(client_id, attrs, headers) do
+    attrs = Map.put(attrs, "consumer_id", get_consumer_id(headers))
+
```
AlexKovalevych: blank line not needed

apps/core/lib/core/oauth/api.ex
```diff
@@ -33,39 +57,34 @@ defmodule Core.OAuth.API do
...
-      case length(data) > 0 do
-        true -> List.first(data)
-        false -> %{}
-      end
+  def get_connection_credentials(client_id, headers) do
+    attrs = %{"consumer_id" => get_consumer_id(headers)}
 
-    fetch_client_credentials({:ok, %{"data" => data}}, id)
+    with {:ok, %{"data" => connections}} <- @mithril_api.get_client_connections(client_id, attrs, headers) do
+      fetch_connection_credentials(connections, client_id)
```
AlexKovalevych: consider return map since it never fails:

```
with {:ok, %{"data" => connections}} <- @mithril_api.get_client_connections(client_id, attrs, headers) do
        {:ok, fetch_connection_credentials(connections, client_id)}
```

apps/ehealth/lib/ehealth/web/proxy.ex
```diff
@@ -25,4 +27,8 @@ defmodule EHealth.Proxy do
 
   def get_proxy_resp_data(%{"error" => error}), do: error
   def get_proxy_resp_data(%{"data" => data}), do: data
+
+  def create_page(paging) do
```
AlexKovalevych: i thinks it's more generic, than a proxy module function

apps/core/lib/core/legal_entities/legal_entities.ex
PavelVesnin: configuration will compiled so charts configuration will not works

apps/core/lib/core/oauth/api.ex
PavelVesnin: agree

apps/ehealth/lib/ehealth/web/proxy.ex
PavelVesnin: `create_page` using just for proxy calls. Another usage is not correct

apps/core/lib/core/legal_entities/legal_entities.ex
AlexKovalevych: Then how does it work in other modules?

apps/ehealth/lib/ehealth/web/proxy.ex
AlexKovalevych: The point is that we want to get rid of proxy calls, since it's not secure - response change on underlying microservice automatically will implicitly change response from the ehealth microservice which is not a desired behaviour. Futhermore `create_page` can be used if you do paging manually (for complex queries, e.g.: capitation)

PavelVesnin: I don't see better place for `create_page` than Proxy module at this time. Possibly `Helpers.Connection`, but this module located in Core (and it's wrong).
Before Proxy it was placed in controller, but all controllers already implemented proxy module.
So it's easy to reuse it in different controllers.
`create_page` cannot be used for manual paging, because it converts string into atoms, when Ecto structs already defined fields as atoms.

apps/core/lib/core/legal_entities/legal_entities.ex
PavelVesnin: it seems that confex updated. Now it works fine, I'll change it

PR #3103, opened 2018-09-12
 GraphQL app base structure

kiurchv:


apps/graphql/lib/graphql_web/resolvers/person.ex
```diff
@@ -0,0 +1,5 @@
+defmodule GraphQLWeb.Resolvers.Person do
+  def list_persons(_parent, _args, _resolution) do
+    {:ok, GraphQL.Person.list_persons()}
```
PavelVesnin: use short alias 
```
alias GraphQL.Person
{:ok, Person.list_persons()}
```

apps/graphql/lib/graphql_web/schema.ex
```diff
@@ -0,0 +1,14 @@
...
+  use Absinthe.Schema
+
+  import_types(GraphQLWeb.Schema.PersonTypes)
+
+  alias GraphQLWeb.Resolvers
+
+  query do
+    @desc "Get all persons"
+    field :persons, list_of(:person) do
+      resolve(&Resolvers.Person.list_persons/3)
```
PavelVesnin: same alias here

AlexKovalevych:
Whats going on?

PavelVesnin:
@AlexKovalevych GraphQL for EHealth admin

AlexKovalevych:
why application name is `GraphQL` is this POC?

PavelVesnin:
Generic name because front-end could use it for admin, cabinet, etc

AlexKovalevych:
what if we need couple different graphql applications?

PavelVesnin:
idea that we have just one endpoint for all graphQL queries for IL.
It's like an `Ehealth` application with REST endpoints

lymychp:
@PavelVesnin how do you plan to combine PRM, MPI, declarations in one endpoint?
spoiler: you cannot connect directly to the mpi and ops DB from ehealth.api))

kiurchv:
@lymychp We can return their JSON API responses in the same way as we do in the EhealthWeb application

PavelVesnin:
@lymychp GraphQL resolver get requested data and decided what to do with request.
You can imagine resolver as controller.
Than we have two options:
 - fetch directly from DB
 - prepare HTTP call

lymychp:
@PavelVesnin @kiurchv  ok than.
Just want to be sure that we all are on the same page.

PavelVesnin:
So for each GraphQL schema will created authorization context and resolver, which clearly understand how fetch and prepare data

Issue #3007, opened 2018-08-28
[IL] Refresh client secret


AlekzandrV:
Implement service according to [logic](https://edenlab.atlassian.net/wiki/spaces/EH/pages/619872633/Manage+more+than+one+client+connections#Managemorethanoneclientconnections-Refreshclientsecret) and [specification](https://uaehealthapi.docs.apiary.io/#reference/public.-medical-service-provider-integration-layer/manage-client-configuration/refresh-client-secret)

child of #2998

AlekzandrV:
Consumer context doesn't work properly

I'm trying to refresh secret for specific connection via proper MIS but receive access_denied error instead of 200

Request
```
curl -X PATCH \
  https://api.dev.asclepius.com.ua/api/clients/41ebae5e-c269-491f-9f78-05bfdae5e397/connections/330c07e5-ae5c-4107-a7c7-1eebd9366d2e/actions/refresh_secret \
  -H 'Authorization: Bearer bTdIMVZjYVhheVpFeVBZekFhaEFwUT09' \
  -H 'Cache-Control: no-cache' \
  -H 'Content-Type: application/json' \
  -H 'Postman-Token: 160ddaf0-c5a9-9774-0592-4ea1a9ce4203' \
  -H 'api-key: UXhEczJacXh3Ri9SUkc4enhHTVVpQT09' \
  -d '{
  "redirect_uri": "http://example5.com"
}'
```

Response
```
{
    "meta": {
        "url": "http://api-svc.il/api/clients/41ebae5e-c269-491f-9f78-05bfdae5e397/connections/330c07e5-ae5c-4107-a7c7-1eebd9366d2e/actions/refresh_secret",
        "type": "object",
        "request_id": "896faa7b-cb2a-4bce-ba1c-0de3061cb63d#1940",
        "code": 401
    },
    "error": {
        "type": "access_denied"
    }
}
```

PavelVesnin:
The reason of invalid behavior is that Mithril does not get `api-key` header in request.
It should be fixed via kong konfiguration

PavelVesnin:
@AlekzandrV deployed and fixed

PR #3145, opened 2018-09-19
do not remove api-key header for requests to Mithril

PavelVesnin:
Header `api-key` was in filtered headers for log security reason, but in fact it blocks correct behavior that expected for [task](https://github.com/edenlabllc/ehealth.api/issues/3007).
Header `api-key` also as another sensitive data should be removed from logs in another way.


AlexKovalevych:
`api-key` should not be logged, please remove `api-key` from logging in this PR

PR #3175, opened 2018-09-25
Unzr#3133

marinakr:


apps/core/lib/core/persons/v1/validator.ex
```diff
@@ -121,15 +122,39 @@ defmodule Core.Persons.Validator do
     end
   end
 
-  defp validate_confidant_persons(%{"confidant_person" => [_ | _] = confidant_persons} = person) do
+  def validate_person_passports(%{} = person) do
+    documents = Map.get(person, "documents", [])
```
BatrakM: As soon as we use document only in the pipe below, maybe it would be better to merge them into one pipe

PR #3180, opened 2018-09-27
 add context plug to GraphQL app

kiurchv:


apps/graphql/lib/graphql_web/authz/schema.ex
```diff
@@ -1,7 +1,7 @@
 defmodule GraphQLWeb.Authz.Schema do
   defmacro __using__(opts \\ []) do
-    meta_key = Keyword.get(opts, :meta_key, :scopes)
-    context_key = Keyword.get(opts, :context_key, :scopes)
+    meta_key = Keyword.get(opts, :meta_key, :scope)
+    context_key = Keyword.get(opts, :context_key, :scope)
```
PavelVesnin: `scope` is correct 👍

apps/graphql/lib/graphql_web/endpoint.ex
```diff
@@ -1,6 +1,10 @@
 defmodule GraphQLWeb.Endpoint do
   use Phoenix.Endpoint, otp_app: :graphql
 
+  alias GraphQLWeb.Plugs.Context
+
+  @scope_header "x-consumer-scope"
```
PavelVesnin: in case when another header will defined here, more readable and easy to group module attributes when in named as `header_scope`, `header_consumer` etc.

apps/graphql/lib/graphql_web/schema.ex
```diff
@@ -10,7 +10,7 @@ defmodule GraphQLWeb.Schema do
   query do
     @desc "Get all persons"
     field :persons, list_of(:person) do
-      #      meta(:scopes, ["person:list"])
+      # meta(:scope, ["person:list"])
```
PavelVesnin: remove or uncomment

apps/graphql/test/web/middleware/authz_middleware_test.exs
```diff
@@ -0,0 +1,54 @@
...
+      }
+    }
+  """
+
+  describe "authorization" do
+    test "success with given scope which includes requested scope", %{conn: conn} do
+      resp_body =
+        conn
+        |> put_scope("legal_entity:list")
+        |> post("/", %{query: @query})
```
PavelVesnin: endpoint path is not important for call, but anyway let's use one approach for path.
Default way for ExUnit it's using RouterHelper for building URL path.
So I propose to emulate this behavior by using Conn.Case function `graphql_path`

apps/graphql/test/web/middleware/authz_middleware_test.exs
```diff
@@ -0,0 +1,54 @@
+defmodule GraphQLWeb.AuthzMiddlewareTest do
```
PavelVesnin: change Authz to full name

apps/graphql/test/web/middleware/authz_middleware_test.exs
kiurchv: I do not think that we need the function for this. How about to keep path into module attribute?

PavelVesnin: than every test modules should have this attribute, when it could defined just in one place - ConnCase

kiurchv: I mean this

PavelVesnin: you can't get module attribute from another module. It should be a function

apps/graphql/lib/graphql_web/endpoint.ex
kiurchv: This naming manner makes hard to understand the attribute purpose: `header_scope` can be misunderstood as it holds the scope of the header and `header_consumer` is the consumer of the header, etc. I prefer to name variables according to [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation)

PR #3188, opened 2018-10-01
LegalEntity schema

PavelVesnin:


apps/graphql/lib/graphql_web/middleware/map_get.ex
```diff
@@ -20,10 +20,11 @@ defmodule GraphQLWeb.Middleware.MapGet do
   end
 
   def call(%{source: source} = resolution, key) do
-    value = case Map.has_key?(source, key) do
-      true -> Map.get(source, key)
-      false -> Map.get(source, to_string(key))
-    end
+    value =
+      case Map.has_key?(source, key) do
```
kiurchv: Maybe `if` fits better?

apps/graphql/lib/graphql_web/schema.ex
```diff
@@ -2,16 +2,19 @@ defmodule GraphQLWeb.Schema do
   @moduledoc false
 
   use Absinthe.Schema
-  use Absinthe.Relay.Schema, :modern
+  use Absinthe.Relay.Schema.Notation, :modern
```
kiurchv: `Absinthe.Relay.Schema` requires `Absinthe.Relay.Schema.Notation`, why do not use it?

kiurchv: So, we should use `Absinthe.Relay.Schema.Notation` in the `LegalEntityTypes` module. Instead, we get duplicated types when we use `Absinthe.Relay.Schema` in another module with type definitions.

PavelVesnin: right, it's better option

Issue #3092, opened 2018-09-11
do not deduplicate children on-line (temp)


svetavedmed:
Temporary on-line deduplication for children should be switched off for both api. 
api/declaration_requests

Changes:
if person age <14 do not save mpi_id

PR #3202, opened 2018-10-01
Add logic for putting mpi_id only for children on declaration_request create

progsmile:
#3092

### DIFF
--- a/apps/core/lib/core/declaration_requests/api/v1/creator.ex
+++ b/apps/core/lib/core/declaration_requests/api/v1/creator.ex
@@ -704,15 +704,15 @@ defmodule Core.DeclarationRequests.API.V1.Creator do
 
   defp put_mpi_id(changeset, person_id, person_birth_date) do
     if passport_age?(person_birth_date) do
-      changeset
-    else
       put_change(changeset, :mpi_id, person_id)
+    else
+      changeset
     end
   end
 
   defp passport_age?(birth_date) do
     case Date.from_iso8601(birth_date) do
-      {:ok, birth_date} -> Timex.diff(Timex.now(), birth_date, :years) < @years_to_have_passport
+      {:ok, birth_date} -> Timex.diff(Timex.now(), birth_date, :years) >= @years_to_have_passport
       _ -> false
     end
   end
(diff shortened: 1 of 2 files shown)


apps/core/lib/core/declaration_requests/api/v1/creator.ex
```diff
@@ -691,12 +694,27 @@ defmodule Core.DeclarationRequests.API.V1.Creator do
...
+    if child?(person_birth_date) do
+      changeset
+    else
+      put_change(changeset, :mpi_id, person_id)
+    end
+  end
+
+  defp child?(birth_date) do
+    case Date.from_iso8601(birth_date) do
+      {:ok, birth_date} -> Timex.diff(Timex.now(), birth_date, :years) < 14
```
imax-iva: does it make sense to put a variable in the configuration?

apps/core/lib/core/declaration_requests/api/v1/creator.ex
```diff
@@ -691,12 +696,27 @@ defmodule Core.DeclarationRequests.API.V1.Creator do
...
+    if passport_age?(person_birth_date) do
+      changeset
+    else
+      put_change(changeset, :mpi_id, person_id)
+    end
+  end
+
+  defp passport_age?(birth_date) do
+    case Date.from_iso8601(birth_date) do
+      {:ok, birth_date} -> Timex.diff(Timex.now(), birth_date, :years) < @years_to_have_passport
```
imax-iva: it seems there is a need to reverse condition:
Timex.diff(Timex.now(), birth_date, :years) > @years_to_have_passport
and change logic in put_mpi_id function

apps/core/lib/core/declaration_requests/api/v1/creator.ex
```diff
@@ -42,6 +42,8 @@ defmodule Core.DeclarationRequests.API.V1.Creator do
   @status_new DeclarationRequest.status(:new)
   @status_approved DeclarationRequest.status(:approved)
 
+  @years_to_have_passport 14
```
imax-iva: there is a sense to use config variable instead of hardcoded value

progsmile: across code

PR #3235, opened 2018-10-05
Fix children search with optional unzr

progsmile:
#3092


apps/core/lib/core/declaration_requests/api/v1/persons.ex
```diff
@@ -26,12 +26,13 @@ defmodule Core.DeclarationRequests.API.Persons do
...
       end
 
+    search_params = maybe_put(search_params, "unzr", person_data["unzr"])
+
     if search_params == %{} do
       {:error, :ignore}
     else
       search_params
       |> Map.put("status", @person_active)
-      |> maybe_put("unzr", person_data["unzr"])
```
marinakr: 1. Unzr should be present here
2. When "maybe_put" function was added? Please choose correct name for it or fix logic

progsmile: 1. It is added when present https://github.com/edenlabllc/ehealth.api/pull/3235/files#diff-c569fc21bcd22a51fcb50da436c58c9dR29
2. Supply better name

progsmile: everywhere you see `maybe` deals with nil

progsmile: If `unzr` exists, `true` in `cond` will evaluate, which add `first_name`, `last_name`, `birth_date` params to search, while only `unzr` and `status` are appropriate
Example: age = 13, unzr = true

Issue #3167, opened 2018-09-24
Enumerable not implemented error 500 GET /divisions


lymychp:
```
  "phoenix": {
      "request_id": "eeafffdf-6477-4443-9b5e-0b705c9d904c#224107",
      "message": "** (Protocol.UndefinedError) protocol Enumerable not implemented for #Ecto.Association.NotLoaded<association :addresses is not loaded>. This protocol is implemented for: DBConnection.PrepareStream, DBConnection.Stream, Date.Range, Ecto.Adapters.SQL.Stream, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, List, Map, MapSet, Postgrex.Stream, Range, Scrivener.Page, Stream, Timex.Interval\n    (elixir) lib/enum.ex:1: Enumerable.impl_for!/1\n    (elixir) lib/enum.ex:141: Enumerable.reduce/3\n    (elixir) lib/enum.ex:1911: Enum.map/2\n    (ehealth) lib/ehealth/web/views/division_view.ex:31: EHealth.Web.DivisionView.render/2\n    (elixir) lib/enum.ex:1294: Enum.\"-map/2-lists^map/1-0-\"/2\n    (phoenix) lib/phoenix/view.ex:332: Phoenix.View.render_to_iodata/3\n    (phoenix) lib/phoenix/controller.ex:740: Phoenix.Controller.do_render/4\n    (ehealth) lib/ehealth/web/controllers/division_controller.ex:1: EHealth.Web.DivisionController.action/2\n",
      "log_type": "error"
    },
    "@timestamp": "2018-09-24T13:44:04.732255938+00:00",
    "tag": "phoenix.error"
  },
```

jeffhsta:
I wasn't able to test it, I think because of some missing plugin in Postgres, but it may fix the issue.

AlekzandrV:
Tested with different scenarios. I can't reproduce this bug

PR #3258, opened 2018-10-10
Preload :address in divisions list

jeffhsta:
I wasn't able to test my change, but I think it should fix it

Fixes https://github.com/edenlabllc/ehealth.api/issues/3167

### DIFF
--- a/apps/core/lib/core/divisions/divisions.ex
+++ b/apps/core/lib/core/divisions/divisions.ex
@@ -54,9 +54,12 @@ defmodule Core.Divisions do
   @default_mountain_group "0"
 
   def list(params) do
-    %Search{}
-    |> changeset(params)
-    |> search(params, Division)
+    page =
+      %Search{}
+      |> changeset(params)
+      |> search(params, Division)
+
+    %{page | entities: preload_address(page.entities)}
   end
 
   def get_by_id!(id) do
@@ -309,5 +312,9 @@ defmodule Core.Divisions do
   defp lowercase_email(params) do
     email = Map.get(params, "email")
     Map.put(params, email, Sanitizer.sanitize(email))
+  end
+
+  defp preload_address(divisions) do
+    PRMRepo.preload(divisions, :addresses)
   end
 end


apps/core/lib/core/divisions/divisions.ex
```diff
@@ -54,9 +54,12 @@ defmodule Core.Divisions do
...
   def list(params) do
-    %Search{}
-    |> changeset(params)
-    |> search(params, Division)
+    page =
+      %Search{}
+      |> changeset(params)
+      |> search(params, Division)
+
+    %{page | entities: preload_address(page.entities)}
```
marinakr: addresses already preloaded, line 236

AlexKovalevych: should be `page.entries`

jeffhsta:
@AlexKovalevych good catch, I fixed it, hope the tests will pass

apps/core/lib/core/divisions/divisions.ex
```diff
@@ -54,9 +54,12 @@ defmodule Core.Divisions do
...
   def list(params) do
-    %Search{}
-    |> changeset(params)
-    |> search(params, Division)
+    page =
+      %Search{}
+      |> changeset(params)
+      |> search(params, Division)
+
+    %{page | entries: preload_address(page.entries)}
```
AlexKovalevych: this won't handle case when changeset is invalid

apps/core/lib/core/divisions/divisions.ex
```diff
@@ -310,4 +313,8 @@ defmodule Core.Divisions do
     email = Map.get(params, "email")
     Map.put(params, email, Sanitizer.sanitize(email))
   end
+
+  defp preload_address(divisions) do
+    PRMRepo.preload(divisions, :addresses)
```
AlexKovalevych: preload should be done at the query level, otherwise it will add +50 queries to db

jeffhsta: I understand that could get the request slow and will consume more resource of the database, but in the current approach here is basically impossible, because there is a common module `Core.Search` that has the `search` function that gets a query dynamically.

What would be the idea case here then?
Should the get divisions logic don't use this `Core.Search.search` function?

AlexKovalevych:
According to this commit the issue was fixed a month ago https://github.com/edenlabllc/ehealth.api/commit/8c60441c38e991e87fb017faa69553fe8047679d
Could you please confirm (maybe by improving the test) that it works as expected?

PR #3264, opened 2018-10-12
EHealth-3113: create related legal entities

PavelVesnin:


apps/graphql/config/prod.exs
```diff
@@ -8,6 +8,7 @@ config :graphql, GraphQLWeb.Endpoint,
     port: {:system, "PORT", "80"}
   ],
   secret_key_base: {:system, "SECRET_KEY"},
+  debug_errors: false,
```
kiurchv: We don't need this since we have same config entry in the global config

PR #3266, opened 2018-10-12
Moved order initialization to Absinthe middleware

BatrakM:


apps/graphql/lib/graphql_web/resolvers/legal_entity.ex
```diff
@@ -8,12 +8,10 @@ defmodule GraphQLWeb.Resolvers.LegalEntity do
   alias Core.LegalEntities.LegalEntity
   alias Core.PRMRepo
 
-  @order_by_regex ~r/(\w+)_(asc|desc)$/
-
-  def list_legal_entities(args, _context) do
+  def list_legal_entities(%{order_by: order_by} = args, _context) do
```
PavelVesnin: what will happens when legalEntities list requested without order_by?

PR #3347, opened 2018-10-25
add contract request queries

kiurchv:


apps/graphql/lib/graphql_web/schema.ex
```diff
@@ -15,16 +15,20 @@ defmodule GraphQLWeb.Schema do
...
   }
 
+  alias Core.ContractRequests.ContractRequest
+  alias Core.Divisions.Division
+  alias Core.Employees.Employee
   alias Core.LegalEntities.LegalEntity
   alias Core.LegalEntities.RelatedLegalEntity
   alias Core.Persons.Person
-  alias GraphQLWeb.Loaders.PRM
+  alias GraphQLWeb.Loaders.{IL, PRM}
```
PavelVesnin: by convention we are using full name aliases

PR #3353, opened 2018-10-25
GraphQL: small code improve

PavelVesnin:


apps/graphql/lib/graphql_web/schema.ex
```diff
@@ -21,7 +21,8 @@ defmodule GraphQLWeb.Schema do
   alias Core.LegalEntities.LegalEntity
   alias Core.LegalEntities.RelatedLegalEntity
   alias Core.Persons.Person
-  alias GraphQLWeb.Loaders.{IL, PRM}
```
kiurchv: I am opposed to the way in which our team is ignoring the Elixir features which improve code readability. I think that refactoring module names referenced by multiple alias/import/require/use directives not so complicated that we should deny this feature. We can easily refactor those module names with find and replace with the regular expression, furthermore, this issue will be completely gone when JakeBecker/elixir-ls#22 will be resolved

PavelVesnin:
not actual anymore

PR #3365, opened 2018-10-29
add date interval scalar type

kiurchv:


apps/graphql/lib/graphql_web/schema/contract_request_types.ex
```diff
@@ -56,6 +56,8 @@ defmodule GraphQLWeb.Schema.ContractRequestTypes do
     field(:edrpou, :string)
     field(:contract_number, :string)
     field(:status, :contract_request_status)
+    field(:start_date, :date_range)
```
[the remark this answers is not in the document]
progsmile: I see it's ok: `#{to_string(today)}/..`

apps/graphql/lib/graphql_web/schema/scalar_types.ex
```diff
@@ -0,0 +1,74 @@
...
+  use Absinthe.Schema.Notation
+
+  alias Absinthe.Blueprint.Input
+
+  scalar :date_range do
+    serialize(&serialize_date_range/1)
+    parse(&parse_date_range/1)
+  end
+
+  defp serialize_date_range(%{first: first, last: last}) when not is_nil(first) and not is_nil(last) do
```
progsmile: Suppose it's better
```
defp serialize_date_range(%{} = data_range) do
    with {:ok, first} <- parse_date(data_range[:first]),
         {:ok, last} <- parse_date(data_range[:last]),
         {:ok, serialized_date_range} <- do_serialize_date_range(first, last) do
      {:ok, serialized_date_range}
    else
      _ -> :error
    end
  end

  defp parse_date(nil), do: {:ok, nil}
  defp parse_date(date), do: Date.to_iso8601(first)

  defp do_serialize_date_range(nil, nil), do: :error
  defp do_serialize_date_range(first, last), do: {:ok, Enum.join([first || "..", last || ".."], "/")}
```

progsmile: I would do the same with `parse_date_range`
@PavelVesnin @kiurchv thoughts?

apps/graphql/test/web/resolvers/contract_request_resolver_test.exs
```diff
@@ -108,6 +108,54 @@ defmodule GraphQLWeb.ContractRequestResolverTest do
...
+      assert 1 == length(resp_entities)
+      assert to_string(today) == hd(resp_entities)["startDate"]
+    end
+
+    test "filter by open date range", %{conn: conn} do
+      nhs()
+
+      today = Date.utc_today()
+
+      for start_date <- [today, Date.add(today, -30)], do: insert(:il, :contract_request, %{start_date: start_date})
```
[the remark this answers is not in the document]
kiurchv: I don't see any complexity in the batch inserts. Actually, I think that grouping similar logic it’s better for understanding than repeating it multiple times

apps/graphql/lib/graphql_web/schema/scalar_types.ex
kiurchv: Agree with this

apps/graphql/test/web/resolvers/contract_request_resolver_test.exs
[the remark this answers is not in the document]
progsmile: ok, notice way you read it (right-left, left-right)

apps/graphql/lib/graphql_web/schema/scalar_types.ex
```diff
@@ -0,0 +1,45 @@
...
+
+  scalar :date_range do
+    serialize(&serialize_date_range/1)
+    parse(&parse_date_range/1)
+  end
+
+  defp serialize_date_range(%{first: first, last: last}) do
+    with {:ok, first} <- serialize_date_range_boundary(first),
+         {:ok, last} <- serialize_date_range_boundary(last) do
+      {:ok, "#{first}/#{last}"}
```
progsmile: is `../..` valid?

kiurchv: Yes, but we will not use it in the DB query in that case

PR #3381, opened 2018-10-30
add assignee_name filter for ContractRequest list field

kiurchv:


apps/graphql/lib/graphql_web/resolvers/contract_request.ex
```diff
@@ -24,6 +25,39 @@ defmodule GraphQLWeb.Resolvers.ContractRequest do
...
+    |> order_by(^order_by)
+  end
+
+  defp prepare_filter([]), do: []
+
+  defp prepare_filter([{:assignee_name, value} | tail]) do
+    assignee_ids =
+      Employee
+      |> join(:inner, [e], p in assoc(e, :party))
+      |> where([e], e.employee_type == "NHS")
```
progsmile: Does only NHS can perform search?

apps/graphql/lib/graphql_web/resolvers/contract_request.ex
```diff
@@ -24,6 +25,39 @@ defmodule GraphQLWeb.Resolvers.ContractRequest do
...
+      |> where(
+        [..., p],
+        fragment(
+          "to_tsvector(concat_ws(' ', ?, ?, ?)) @@ to_tsquery(?)",
```
progsmile: Why `p.last_name` isn't first among `first_name`, `second_name` params? Is it better than `like` search?

kiurchv: In the current case, word order makes not so much sense. All lexemes in the `tsvector` will have the same priority, but we can place `last_name` first  assuming that most of the queries will start from last name

apps/graphql/lib/graphql_web/resolvers/contract_request.ex
kiurchv: Nope, but we can assign only NHS employee to contract request

PR #3390, opened 2018-10-31
GraphQL: approve contract request

PavelVesnin:


apps/core/lib/core/contract_requests/contract_requests.ex
```diff
@@ -1294,9 +1296,8 @@ defmodule Core.ContractRequests do
     end
   end
 
-  def get_by_id(id) do
-    Repo.get(ContractRequest, id)
-  end
+  def get_by_id(id), do: Repo.get(ContractRequest, id)
```
AlexKovalevych: since these are separate functions, blank line is required - otherwise it looks like it's the same function

apps/core/lib/core/contract_requests/contract_requests.ex
```diff
@@ -1359,6 +1360,15 @@ defmodule Core.ContractRequests do
     end
   end
 
+  defp validate_contract_request_id(nil, _), do: :ok
+
+  defp validate_contract_request_id(id_from_request, id_from_signed_content)
```
AlexKovalevych: Let's use pattern matching, e.g.:

```
defp validate_contract_request_id(nil, _), do: :ok
defp validate_contract_request_id(id, id), do: :ok
defp validate_contract_request_id(_id_from_request, _id_from_signed_content) do
  {:error, {:bad_request, "Contract request id doesn't match with id in signed content"}}
end
```

apps/core/lib/core/contract_requests/render.ex
```diff
@@ -0,0 +1,153 @@
+defmodule Core.ContractRequests.Render do
+  @moduledoc """
+  Contract Request should know how to render itself,
+  because it's important for Digital Signature validation
+  and validation can be used in different applications
+  """
+
+  def render(contract_request, references) do
```
AlexKovalevych: Probably we need type hinting for `contract_request` here, since you get atoms from the struct

apps/core/lib/core/contract_requests/render.ex
```diff
@@ -0,0 +1,153 @@
+defmodule Core.ContractRequests.Render do
```
AlexKovalevych: This is actually a `renderer`, not `render`

apps/core/lib/core/contract_requests/contract_requests.ex
PavelVesnin: don't understand what you mean. It's already with pattern matching

AlexKovalevych: compare 

```
  defp validate_contract_request_id(id_from_request, id_from_signed_content)
       when id_from_request == id_from_signed_content,
       do: :ok
```

with 

```
defp validate_contract_request_id(id, id), do: :ok
```

PavelVesnin: holy crap, is that works?

AlexKovalevych: yup )

Issue #3117, opened 2018-09-13
Legal entity nhsVerify/deactivate


danakonovalenko:
NHS admin must be able to verify or deactivate legal entity according to [scheme](
https://github.com/edenlabllc/ehealth.web/blob/master/packages/mock-server/__schema__/legal_entities.graphql) and [logic verify](https://edenlab.atlassian.net/wiki/x/iewr) and [logic deactivate](https://edenlab.atlassian.net/wiki/x/-m8n)

```
- nhsVerifyLegalEntity(id: ID!): LegalEntity
- deactivateLegalEntity(id: ID!): LegalEntity
```

-  [x] [test cases](https://edenlab.atlassian.net/wiki/spaces/EH/pages/3443558/Legal+entity+tests)

danakonovalenko:
to Verify LegalEntity it must be in status 'ACTIVE'. Closed legal entity can't be verified.
As for now it's possible to Verify deactivated legal entity

PR #3393, opened 2018-11-01
Dev 3117 legal entity nhs verify deactivate

progsmile:
#3117


apps/core/lib/core/employees/employee_updater.ex
```diff
@@ -28,13 +28,16 @@ defmodule Core.Employees.EmployeeUpdater do
...
-    with employee <- Employees.get_by_id!(employee_id),
+    with %Employee{} = employee <- Employees.get_by_id(employee_id),
          :ok <- check_legal_entity_id(legal_entity_id, employee),
          :ok <- check_transition(employee, with_owner),
          active_employees <- get_active_employees(employee),
          :ok <- revoke_user_auth_data(employee, active_employees, headers),
          {:ok, _} <- @ops_api.terminate_employee_declarations(employee_id, user_id, reason, "", headers) do
       set_employee_status_as_dismissed(employee, headers)
+    else
+      nil -> {:error, {:not_found, "Employee not found"}}
```
PavelVesnin: are you sure that response with `nil` can be only for `Employees.get_by_id`?
Suggest you to make function: 
```
def get_employee(id) do
  case Employees.get_by_id(id) do
     %Employee{} = employee -> {:ok, employee}
     nil -> {:error, {:not_found, "Employee not found"}}
  end
end
```

progsmile: great idea

apps/core/lib/core/legal_entities/legal_entity_updater.ex
```diff
@@ -17,10 +17,13 @@ defmodule Core.LegalEntities.LegalEntityUpdater do
...
 
   def deactivate(id, headers) do
-    with legal_entity <- LegalEntities.get_by_id!(id),
+    with %LegalEntity{} = legal_entity <- LegalEntities.get_by_id(id),
          :ok <- check_transition(legal_entity),
          :ok <- deactivate_employees(legal_entity, headers) do
       update_legal_entity_status(legal_entity, headers)
+    else
+      nil -> {:error, {:not_found, "LegalEntity not found"}}
+      error -> error
```
PavelVesnin: I see 4 places where you catch `nil`.
Why you can't rescue exception instead of duplicating code?

PavelVesnin: yes, but in this case you will duplicate a lot of code and use additional `else` in every `with`.
It's also not so good. 
I'd like to find more elegant and generic way to handle this case

progsmile: LegalEntities.get_or_fail ?

apps/core/lib/core/legal_entities/legal_entities.ex
```diff
@@ -128,6 +128,15 @@ defmodule Core.LegalEntities do
     |> PRMRepo.one()
   end
 
+  def fetch_by_id(id) do
```
PavelVesnin:
```suggestion
  def fetch_by_id(id) do
       case get_by_id(id) do
           %LegalEntity{} = legal_entity -> {:ok, legal_entity}
           _ -> {:error, {:not_found, "LegalEntity not found"}}
        end
    end
```

PavelVesnin: inline case more readable

PR #3402, opened 2018-11-02
Check legal_entity to be active on verify

progsmile:
#3117


apps/core/lib/core/legal_entities/legal_entities.ex
```diff
@@ -253,10 +253,12 @@ defmodule Core.LegalEntities do
     {:error, {:conflict, "LegalEntity is VERIFIED and cannot be VERIFIED."}}
   end
 
-  defp check_nhs_verify_transition(%LegalEntity{nhs_verified: false}), do: :ok
-
-  defp check_nhs_verify_transition(_) do
-    {:error, {:conflict, "LegalEntity is VERIFIED and cannot be VERIFIED."}}
+  defp check_nhs_verify_transition(%LegalEntity{nhs_verified: nhs_verified, status: status}) do
+    cond do
```
kiurchv: Maybe it is better to use own function for each check?

progsmile: yep, it's better way

PR #3414, opened 2018-11-06
error text

marinakr:


apps/core/lib/core/legal_entities/validator.ex
```diff
@@ -106,7 +106,7 @@ defmodule Core.LegalEntities.Validator do
     |> cast(%{"edrpou" => content_edrpou(content)}, Map.keys(types))
     |> validate_required(Map.keys(types))
     |> validate_format(:edrpou, ~r/^[0-9]{8,10}$/)
-    |> validate_inclusion(:edrpou, [edrpou], message: "EDRPOU is not match with signer drfo")
+    |> validate_inclusion(:edrpou, [edrpou], message: "EDRPOU is not match with legal_entity edrpou")
```
AlexKovalevych: `EDRPOU doesn't not match legal_entity edrpou`

PR #3418, opened 2018-11-06
Using phone number as auth method is now limited by global parameter

BatrakM:


apps/core/lib/core/declaration_requests/api/v2/mpi_search.ex
```diff
@@ -7,19 +7,26 @@ defmodule Core.DeclarationRequests.API.V2.MpiSearch do
...
+    |> @mpi_api.search([])
+    |> search_result(:all)
+  end
+
   def search(person, headers \\ []) do
     with {:ok, search_params} <- Persons.get_search_params(person) do
       search_params
       |> @mpi_api.search(headers)
-      |> search_result()
+      |> search_result(:single)
```
AlexKovalevych: suggested naming: `:one`

apps/core/lib/core/declaration_requests/api/v2/creator.ex
```diff
@@ -56,24 +56,58 @@ defmodule Core.DeclarationRequests.API.V2.Creator do
...
+    search_params =
+      changeset
+      |> get_field(:data)
+      |> get_in(["person", "authentication_methods"])
+      |> Enum.find(fn authentication_method -> Map.has_key?(authentication_method, "phone_number") end)
+      |> prepare_auth_phone_number_param()
+
+    with %{"auth_phone_number" => _} <- search_params,
+         {:ok, persons} <- mpi_search(search_params) do
+      case Enum.count(persons) < max_phone_number do
```
AlexKovalevych: `if` statement is better, since you have only 2 cases

apps/core/lib/core/declaration_requests/api/v2/creator.ex
```diff
@@ -56,24 +56,58 @@ defmodule Core.DeclarationRequests.API.V2.Creator do
   defp insert_declaration_request(params, user_id, auxiliary_entities, headers) do
     params
     |> changeset(user_id, auxiliary_entities, headers)
-    |> determine_auth_method_for_mpi(params["channel"], auxiliary_entities[:person_id])
+    |> determine_auth_method_for_mpi(params["channel"], auxiliary_entities)
     |> generate_printout_form(auxiliary_entities[:employee])
     |> do_insert_declaration_request()
   end
 
+  defp prepare_auth_phone_number_param(%{"phone_number" => phone_number}), do: %{"auth_phone_number" => phone_number}
```
AlexKovalevych: what is the reason to rename parameter from `phone_number` to `auth_phone_number`?

BatrakM: Search by phone_number is already implemented, so i had to use another name for parameter

apps/core/config/config.exs
```diff
@@ -312,6 +312,9 @@ config :core, Core.Guardian,
 # Deviation koeficient 0..1, equal to percents
 config :core, Core.MedicationDispense.API, deviation: {:system, :float, "DEVIATION", 0.1}
 
+config :core, Core.DeclarationRequests.API.V1.Creator,
+  use_phone_number_auth_limit: {:system, "USE_PHONE_NUMBER_AUTH_LIMIT", true}
```
AlexKovalevych: `{:system, :boolean, "USE_PHONE_NUMBER_AUTH_LIMIT", true}`

BatrakM: done

PR #3462, opened 2018-11-08
GraphQL: fixes and improves for contract details

PavelVesnin:


apps/core/lib/core/contracts/contract.ex
```diff
@@ -3,6 +3,7 @@ defmodule Core.Contracts.Contract do
 
   use Ecto.Schema
 
+  alias Core.ContractRequests.ContractRequest
```
AlexKovalevych: why do you need this alias here?

PavelVesnin: mistake

PR #3485, opened 2018-11-11
added databaseId map for PRM loader

PavelVesnin:


PavelVesnin:
duplicated with #3462

PR #3508, opened 2018-11-12
Add review content fields to ContractRequest

kiurchv:


apps/graphql/test/web/resolvers/contract_request_resolver_test.exs
```diff
@@ -612,6 +612,56 @@ defmodule GraphQLWeb.ContractRequestResolverTest do
...
+
+      resp_body =
+        conn
+        |> put_client_id()
+        |> post_query(query, variables)
+        |> json_response(200)
+
+      resp_entity = get_in(resp_body, ~w(data contractRequest))
+
+      assert nil == resp_body["errors"]
```
PavelVesnin: refute

apps/core/lib/core/contract_requests/renderer.ex
```diff
@@ -52,6 +52,16 @@ defmodule Core.ContractRequests.Renderer do
     })
   end
 
+  def render_review_content(%ContractRequest{} = contract_request, associations) do
+    {legal_entity, review_fields} = Map.pop(associations, :contractor_legal_entity)
+    contractor_legal_entity = Map.take(legal_entity, ~w(id edrpou name)a)
+
+    contract_request
+    |> Map.take(~w(id)a)
```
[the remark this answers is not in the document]
kiurchv: Just in case when we need more fields from the `contract_request`. Also, using `Map.take` we are getting the prefilled map in one function call

apps/graphql/test/web/resolvers/contract_request_resolver_test.exs
kiurchv: I'll replace all nil assertions in the separate PR

Issue #3460, opened 2018-11-08
Prolongate contract mutation


danakonovalenko:
Add possibility for NHS admin to update contract end date for merged legal entities using mutation:
```
prolongateContract(input: ProlongateContractInput!): ProlongateContractPayload
```

- [x] [logic](https://edenlab.atlassian.net/wiki/x/c4APK) / [schema](https://github.com/edenlabllc/ehealth.schema/blob/master/__schema__/contracts.graphql)/ design(TBD!)
- [x] implement mutation
- [x] test

danakonovalenko:
please, see comment in [test cases](https://edenlab.atlassian.net/wiki/x/eIBxK)

PR #3511, opened 2018-11-12
Implement prolongate contract on graphql

progsmile:
#3460


apps/core/lib/core/contracts/contracts.ex
```diff
@@ -416,6 +417,13 @@ defmodule Core.Contracts do
     end
   end
 
+  def fetch_by_id(id) do
+    case get_by_id(id) do
+      %Contract{} = contract -> {:ok, contract}
+      _ -> {:error, {:not_found, "Contract not found"}}
+    end
+  end
```
PavelVesnin: it's a generic function that should in boundary module `Contracts`

progsmile: suggestions?

progsmile: matching from previous `fetch_by_id(id, params)` is different
see no problems

Issue #3468, opened 2018-11-09
LE approve new. Rest


svetavedmed:
as NHS admin I want to review legal_entity with tracking statuses of reviewing and approving so that I know if legal_entity is created correctly or should be changed

- [ ] add new fields to DB and save new data to this fields
   - 'review_comment', text
   - nhs_review, boolean
- [ ] change logic on [update legal entity](https://uaehealthapi.docs.apiary.io/#reference/public.-medical-service-provider-integration-layer/legal-entities/create/update-legal-entity)
- [ ] add new fields for MISes
   - [get LE by ID](https://uaehealthapi.docs.apiary.io/#reference/public.-medical-service-provider-integration-layer/legal-entities/get-legal-entities)
   - [create/update legal entity](https://uaehealthapi.docs.apiary.io/#reference/public.-medical-service-provider-integration-layer/legal-entities/get-legal-entity-by-id)

progsmile:
nhs_comment will be added in #3469 there is no functionality for now

PR #3538, opened 2018-11-15
Add nhs_reviewed, nhs_comment legal_entity fields and their logic

progsmile:
#3468


apps/core/lib/core/legal_entities/legal_entities.ex
```diff
@@ -69,6 +69,7 @@ defmodule Core.LegalEntities do
     email
     is_active
     nhs_verified
+    nhs_reviewed
```
PavelVesnin: why field `nhs_comment` not added?

apps/ehealth/test/web/controllers/legal_entity_controller_test.exs
```diff
@@ -664,13 +702,16 @@ defmodule EHealth.Web.LegalEntityControllerTest do
...
-      refute is_nil(resp["data"]["nhs_verified"])
-      refute resp["data"]["nhs_verified"]
+      assert %{"data" => resp_data} =
+               conn
+               |> put_client_id_header(id)
+               |> get(legal_entity_path(conn, :show, id))
+               |> json_response(200)
+
+      assert %{"mis_verified" => "VERIFIED", "nhs_reviewed" => _} = resp_data
+      refute is_nil(resp_data["nhs_verified"])
```
kiurchv:
```suggestion
      assert resp_data["nhs_verified"]
```

apps/ehealth/test/web/controllers/legal_entity_controller_test.exs
```diff
@@ -526,18 +568,14 @@ defmodule EHealth.Web.LegalEntityControllerTest do
       msp()
       %{id: id, edrpou: edrpou} = insert(:prm, :legal_entity)
 
-      resp =
-        conn
-        |> put_client_id_header(id)
-        |> get(legal_entity_path(conn, :index, edrpou: edrpou))
-        |> json_response(200)
+      assert %{"data" => resp_data, "paging" => _} =
```
PavelVesnin: If response will not contain field `paging`, you will get a long error message in which hard to understand what is wrong on comparison.
That's why in previous version was assertion with `Map.has_key?`

apps/core/lib/core/legal_entities/legal_entities.ex
progsmile: there is no logic with it

apps/core/priv/prm_repo/migrations/20181114092659_add_nhs_comment_nhs_reviewed_fields_to_le.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Core.Repo.Migrations.AddNhsCommentNhsReviewedFieldsToLe do
+  use Ecto.Migration
+
+  def change do
+    alter table(:legal_entities) do
+      add(:nhs_reviewed, :boolean, default: false)
+      add(:nhs_comment, :text, default: "")
```
AlexKovalevych: since this field is not used, let's not add it at all

PR #3546, opened 2018-11-15
Split ContractRequests into CapitationContractRequest and ReimbursementContractRequest (STI)

PavelVesnin:


apps/graphql/test/web/resolvers/contract_request_resolver_test.exs
```diff
@@ -728,7 +728,12 @@ defmodule GraphQLWeb.ContractRequestResolverTest do
       printout_content = "<html></html>"
 
       contract_request =
-        insert(:il, :contract_request, status: ContractRequest.status(:new), printout_content: printout_content)
+        insert(
+          :il,
+          :contract_request,
```
progsmile: How about splitting factory for different entities?

PavelVesnin: you're right

PR #3559, opened 2018-11-15
Try to fix Travis DB connection error

PavelVesnin:


apps/ehealth/mix.exs
```diff
@@ -76,7 +76,13 @@ defmodule EHealth.Mixfile do
 
   defp aliases do
     [
-      "ecto.setup": &ecto_setup/1
+      "ecto.setup": &ecto_setup/1,
+      test: [
```
AlexKovalevych: i believe we have such alias in `Core/mix.exs`, why do you duplicate it's here?

AlexKovalevych:
According to first build, it doesn't look like the problem is solved

PR #3560, opened 2018-11-16
updated deps, fixed credo warns

PavelVesnin:


apps/core/mix.exs
```diff
@@ -63,8 +63,8 @@ defmodule Core.MixProject do
         "ecto.create --repo Core.FraudRepo",
         "ecto.create --repo Core.PRMRepo",
         "ecto.create --repo Core.EventManagerRepo",
-        "ecto.migrate",
-        "run priv/repo/seeds.exs"
+        "ecto.migrate"
+        #        "run priv/repo/seeds.exs"
```
AlexKovalevych: probably we can remove it, if we don't need it?

PR #3567, opened 2018-11-16
Deduplicate filtering logic in GraphQL application

kiurchv:


apps/graphql/lib/graphql_web/resolvers/dictionary_resolver.ex
```diff
@@ -23,7 +23,7 @@ defmodule GraphQLWeb.Resolvers.DictionaryResolver do
   end
 
   defp prepare_filter([{:name, value} | tail]) do
-    [{:name, {:ilike, "%#{value}%"}} | prepare_filter(tail)]
+    [{:name, {:like, value}} | prepare_filter(tail)]
```
progsmile: Suppose there would be cases when we need to search from start of string, or with some preffix

progsmile: I left it `:ilike` to mark that's case insensitive search

kiurchv: I have not found any such case yet. But we can change this behavior later if needed

apps/graphql/lib/graphql_web/resolvers/contract_request_resolver.ex
```diff
@@ -2,7 +2,7 @@ defmodule GraphQLWeb.Resolvers.ContractRequestResolver do
   @moduledoc false
 
   import Absinthe.Resolution.Helpers, only: [on_load: 2]
-  import Ecto.Query, only: [where: 2, where: 3, join: 4, select: 3, order_by: 2]
+  import Ecto.Query, only: [where: 3, join: 4, select: 3, order_by: 2]
```
progsmile: Why shouldn't import all? It can be annoying to grab every single function from Query module

apps/graphql/lib/graphql_web/resolvers/dictionary_resolver.ex
kiurchv: I'm going to rewrite all filtering logic, anyway. The goal of this PR is code deduplication

apps/graphql/lib/graphql_web/resolvers/contract_request_resolver.ex
progsmile: I know that's correct to do so, but it's not practical

PR #3598, opened 2018-11-20
Add filtering helper and middleware

kiurchv:


apps/graphql/lib/graphql/helpers/filtering.ex
```diff
@@ -0,0 +1,103 @@
...
+    |> filter(conditions)
+  end
+
+  defp introspect(query, field) do
+    %{from: {_, queryable}} = Ecto.Queryable.to_query(query)
+
+    Enum.reduce_while(
+      ~w(association embed type)a,
+      nil,
+      fn type, _ ->
```
progsmile: What kind of type does it return?

apps/graphql/lib/graphql/helpers/filtering.ex
```diff
@@ -0,0 +1,103 @@
...
+        if type = queryable.__schema__(type, field), do: {:halt, type}, else: {:cont, nil}
```
progsmile: Suppose it matches always

kiurchv: It matches only when we have a certain association or embed, otherwise it will return type of the field

PR #3602, opened 2018-11-20
use proper column name for STI

kiurchv:


PavelVesnin:
I have already implement it in ContractRequest PR

apps/core/lib/core/contract_requests/requests/capitation_contract_request.ex
```diff
@@ -1,17 +1,15 @@
 defmodule Core.ContractRequests.CapitationContractRequest do
   @moduledoc false
 
-  @contract_type "CAPITATION"
+  @inheritance_name "CAPITATION"
```
progsmile: Why it's called @inheritance_name ?

progsmile: it's too infrastructural

kiurchv: Because this column serves only to one purpose: provide a way to map records in the DB to models

kiurchv: `inheritance_column` and `inheritance_name` are both inspired by STI implementations in other ORMs

See similar implementation in ActiveRecord, for example: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/inheritance.rb

PR #3640, opened 2018-11-23
set Employee.type filter as list, added NHS_SIGNER employee type

PavelVesnin:


apps/graphql/test/web/resolvers/employee_resolver_test.exs
```diff
@@ -117,8 +119,11 @@ defmodule GraphQLWeb.EmployeeResolverTest do
       resp_entities = get_in(resp_body, ~w(data employees nodes))
 
       assert nil == resp_body["errors"]
-      assert 1 == length(resp_entities)
-      assert @employee_type_admin == hd(resp_entities)["employeeType"]
+      assert 2 == length(resp_entities)
+
+      Enum.each(resp_entities, fn employee ->
+        assert employee["employeeType"] in employee_types
```
progsmile: are tests ok here?

PavelVesnin: yes, what could be wrong?

Issue #3522, opened 2018-11-13
Get Contract REST


IrinaRevutska:
[confluence Contract List](https://edenlab.atlassian.net/wiki/spaces/EH/pages/529039458/Get+Contracts+List)
[apiary Contract  list](https://uaehealthapi.docs.apiary.io/#reference/public.-contracts/contracts/public.-get-contracts-list)
- create separate routes for capitaion and reimbursement versions of a contract
- add contract_type
- add medical_program_id
- add filtration by medical_program_id

[confluence Contract details](https://edenlab.atlassian.net/wiki/spaces/EH/pages/530481202/Get+Contract+by+ID)
[apiary Contract  details](https://uaehealthapi.docs.apiary.io/#reference/public.-contracts/contracts/public.-get-contract-details)
- create separate routes for capitaion and reimbursement versions of a contract
- add contract_type
- add medical_program_id
- add filtration by medical_program_id

[DataModel](https://edenlab.atlassian.net/wiki/spaces/EH/pages/671252546/Contract+Reimbursment+version+Data+model)

svetavedmed:
- [ ] pharmacy owner should get only reimbursement contracts and only its own. As for now I can get too much

PR #3679, opened 2018-11-27
Implement reimbursement contracts list and show

progsmile:
#3522


apps/core/lib/core/contracts/contracts.ex
```diff
@@ -424,28 +424,36 @@ defmodule Core.Contracts do
     |> PRMRepo.one()
   end
 
-  def get_by_id(id, params) do
```
progsmile: now it's `get_by_id_with_client_validation `

PR #3711, opened 2018-11-29
Split ContractRequest schema

kiurchv:


apps/graphql/features/contexts/context.exs
```diff
@@ -0,0 +1,355 @@
...
+
+    %{conn: ConnTest.build_conn()}
+  end)
+
+  scenario_finalize(fn status, state ->
+    Sandbox.checkin(Repo)
+    Sandbox.checkin(PRMRepo)
+    Sandbox.checkin(EventManagerRepo)
+
+    with {:ok, _} <- status, do: Mox.verify!()
```
progsmile: use if

apps/graphql/features/contexts/context.exs
```diff
@@ -0,0 +1,355 @@
...
+             )
+
+      {:ok, state}
+    end
+  )
+
+  then_(
+    ~r/^I should receive collection with (?<count>\d+) items?$/,
+    fn %{resp_entities: resp_entities} = state, %{count: count} ->
+      assert resp_entities
```
progsmile: What does `resp_entities` contain in not found case? Suppose should receive empty list

apps/graphql/features/contexts/context.exs
```diff
@@ -0,0 +1,355 @@
...
+  then_(
+    ~r/^the (?<field>[\w_]+) of the first item in the collection should be "(?<value>[^"]+)"$/,
+    fn %{resp_entities: resp_entities} = state, %{field: field, value: value} ->
+      resp_value =
+        resp_entities
+        |> hd()
+        |> Map.get(field)
+        |> to_string()
+
+      assert value = resp_value
```
progsmile: use ==

apps/graphql/features/contexts/context.exs
```diff
@@ -0,0 +1,355 @@
...
+  def put_client_id(conn, id \\ UUID.generate()) do
+    metadata = Jason.encode!(%{"client_id" => id})
+    put_req_header(conn, @consumer_metadata_header, metadata)
+  end
+
+  def post_query(conn, query, variables \\ %{}) do
+    post(conn, @graphql_path, %{query: query, variables: variables})
+  end
+
+  def prepare_attrs(queryable, attrs) when is_map(attrs), do: prepare_attrs(queryable, Map.to_list(attrs))
```
progsmile: In codebase most time `params` variable is mentioned, let's use such name

apps/graphql/features/contexts/context.exs
kiurchv: This step definition checks collection presence and length, in the case when we expect to receive no collection at all we use another condition, see here: https://github.com/edenlabllc/ehealth.api/pull/3711/files#diff-fcb81f96426ffd0a0f94a13fdce6af56R239

apps/graphql/lib/graphql_web/schema/contract_request_types.ex
```diff
@@ -181,48 +116,51 @@ defmodule GraphQLWeb.Schema.ContractRequestTypes do
...
 
-  node object(:contract_request) do
+  interface :contract_request do
+    field(:id, non_null(:id))
     field(:database_id, non_null(:id))
     field(:contract_number, :string)
-    field(:parent_contract, :contract, resolve: load_by_parent(PRM, CapitationContract))
-    field(:previous_request, :contract_request, resolve: dataloader(IL))
-    field(:assignee, :employee, resolve: load_by_parent(PRM, Employee))
+    # field(:parent_contract, :contract)
```
progsmile: remove

apps/graphql/features/capitation_contract_requests/get_all.feature
```diff
@@ -0,0 +1,76 @@
...
+
+  Scenario: Request all items with NHS client
+    Given there are 2 capitation contract requests exist
+    And my scope is "contract_request:read"
+    And my client type is "NHS"
+    When I request first 10 capitation contract requests
+    Then no errors should be returned
+    And I should receive collection with 2 items
+
+  Scenario: Request belonging items with MSP client
```
progsmile: 👍👍

apps/graphql/features/contexts/context.exs
kiurchv: We cannot use match operator in `if` there, so I'd decided to use `with` instead

apps/graphql/features/contexts/context.exs
kiurchv: For what purpose?

apps/graphql/features/contexts/context.exs
progsmile: if match?({:ok, _}, status) do ..

apps/graphql/features/contexts/context.exs
kiurchv: Nope, when inserting a new record to DB we are passing its fields (or attributes). Params are related only to HTTP requests

apps/graphql/features/contexts/context.exs
progsmile: to be more explicit

apps/graphql/lib/graphql_web/schema/contract_request_types.ex
kiurchv: No, we still need this field, I'll uncomment this when Contract schema will be ready

apps/graphql/features/contexts/context.exs
kiurchv: Yes, but `with` does the same under the hood

apps/graphql/features/contexts/context.exs
```diff
@@ -0,0 +1,355 @@
...
+      for row <- table_data do
+        attrs = prepare_attrs(LegalEntity, row)
+        insert(:prm, :legal_entity, attrs)
+      end
+
+      {:ok, state}
+    end
+  )
+
+  when_(
```
PavelVesnin: suggest to move queries to another file

kiurchv: I've tried several approaches to organize step definitions better but ended with this one global context file with all step definitions. Currently, development of our BDD testing infrastructure is on the very early stage. For now, I want to write more tests and make its refactoring later.

apps/graphql/lib/graphql_web/schema/contract_request_types.ex
progsmile: what about next line?
# field(:previous_request, :contract_request)

kiurchv:
@danakonovalenko @IrinaRevutska Since this PR introduces BDD approach on eHealth, I'd like to hear your opinions on Gherkin feature files

apps/graphql/lib/graphql_web/schema/contract_request_types.ex
kiurchv: It seems that previous line no more needed, agree

apps/graphql/lib/graphql_web/schema/contract_request_types.ex
```diff
@@ -114,7 +49,7 @@ defmodule GraphQLWeb.Schema.ContractRequestTypes do
 
     payload field(:approve_contract_request) do
       meta(:scope, ~w(contract_request:update))
-      middleware(ParseIDs, id: :contract_request)
+      middleware(ParseIDs, id: :capitation_contract_request)
```
progsmile: Why there is :capitation_contract_request ?

Issue #3457, opened 2018-11-08
Add DRFO validation for Contract Request Signs


danakonovalenko:
To guarantee full matching of data in dataBase and data in Digital Signature we need to compare 
DS.drfo with party.tax_id
Add validation
- [x] for **Create/Approve/Decline**
Check that DRFO in Certificate details is equal to tax_id in party
 - Get parties.tax_id using party_users.party_id by user_id.
 - Compare DRFO in Certificate with party.tax_id 
   ```
     - Convert DRFO and TAX_ID to uppercase
     - Compare DRFO and TAX_ID as Cyrillic letters
     - Convert DRFO to Cyrillic and compare as Cyrillic letters
   ```
 - In case validation fails - generate 422 error

- [x] for **NHS Sign/ MSP Sign**
Check that DRFO in Certificate details is equal to tax_id in party
 - Get `parties.tax_id` using `employee.nhs_signer_id.party_id` or `employee.contractor_onwer_id.party_id`by user_id.
 - Compare DRFO in Certificate with party.tax_id 
    ```
     - Convert DRFO and TAX_ID to uppercase
     - Compare DRFO and TAX_ID as Cyrillic letters
     - Convert DRFO to Cyrillic and compare as Cyrillic letters
   ```
 - In case validation fails - generate 422 error

svetavedmed:
*on Create CR*
- [ ] in case edrpou does not exist in DS i receive an error "description":"DRFO does not match signer drfo","params":["3173108921"]", which is lie as drfo match to the signer drfo and only edrpou should be present. Can we change an error to "EDRPOU does not match legal_entity edrpou" or "edrpou should be present" in order to understand that that is a problem with edrpou only.

PR #3733, opened 2018-11-30
Add drfo validation for cr signs#3457

imax-iva:
https://github.com/edenlabllc/ehealth.api/issues/3457


apps/core/lib/core/contract_requests/contract_requests.ex
```diff
@@ -276,6 +277,7 @@ defmodule Core.ContractRequests do
 
     with :ok <- JsonSchema.validate(:contract_request_sign, params),
          {:ok, %{"content" => content, "signers" => [signer]}} <- decode_signed_content(params, headers),
+         :ok <- SignatureValidator.check_drfo(signer, user_id, "contract_request_approve"),
```
PavelVesnin: why this validation without `path` parameter, such as `"$.drfo"` ?

imax-iva: SignatureValidator.check_drfo has different arity and logic

PR #3742, opened 2018-12-03
fix year condition + remove deduplication

marinakr:


apps/core/lib/core/contract_requests/validator.ex
```diff
@@ -669,9 +669,10 @@ defmodule Core.ContractRequests.Validator do
   def validate_end_date(%{"start_date" => start_date, "end_date" => end_date}) do
     start_date = Date.from_iso8601!(start_date)
     end_date = Date.from_iso8601!(end_date)
+    days_in_year = if Date.leap_year?(Date.utc_today()), do: 366, else: 365
```
AlexKovalevych: either handle all leap year cases or non of them

PR #3758, opened 2018-12-04
Changed status_reason column type in contracts and contract requests

BatrakM:


apps/ehealth/test/web/controllers/contract_controller_test.exs
```diff
@@ -927,6 +927,65 @@ defmodule EHealth.Web.ContractControllerTest do
...
+      nhs()
+      contract = insert(:prm, :capitation_contract)
+      params = %{"status_reason" => String.duplicate("a", 3001)}
+
+      resp =
+        conn
+        |> put_client_id_header(contract.nhs_legal_entity_id)
+        |> patch(contract_path(conn, :terminate, @capitation, contract.id), params)
+
+      assert resp = json_response(resp, 422)
```
PavelVesnin: `json_response` already assert status code, so you don't need assertion here. You can use pipeline here

```
resp =
        conn
        |> put_client_id_header(contract.nhs_legal_entity_id)
        |> patch(contract_path(conn, :terminate, @capitation, contract.id), params)
        |> json_response(422)
```

apps/ehealth/test/web/controllers/contract_controller_test.exs
```diff
@@ -927,6 +927,65 @@ defmodule EHealth.Web.ContractControllerTest do
...
+      nhs()
+      contract = insert(:prm, :capitation_contract)
+      params = %{"status_reason" => "%"}
+
+      resp =
+        conn
+        |> put_client_id_header(contract.nhs_legal_entity_id)
+        |> patch(contract_path(conn, :terminate, @capitation, contract.id), params)
+
+      assert resp = json_response(resp, 422)
```
PavelVesnin: same here

Issue #3756, opened 2018-12-04
Additional validation errors (extend error message)


lymychp:
- [ ] test for Changeset validation error with full error message fields (rules, path, description)
- [ ]  test for JSON Schema validation error with full error message fields (rules, path, description)

PavelVesnin:
[ ] test for Changeset validation error with full error message fields (rules, path, description)
[ ] test for JSON Schema validation error with full error message fields (rules, path, description)

PR #3765, opened 2018-12-05
Format errors on graphql

progsmile:
#3756


apps/graphql/test/web/resolvers/legal_entity_merge_job_resolver_test.exs
```diff
@@ -15,6 +15,8 @@ defmodule GraphQLWeb.LegalEntityMergeJobResolverTest do
 
   setup :verify_on_exit!
 
+  @moduletag :wipp
```
PavelVesnin: do we need it?

progsmile: no

Issue #3469, opened 2018-11-09
LE approve new. GraphQL


svetavedmed:
- [x] add mutations
   - nhs_reviewed = true
   - nhs_verified = false
   - comment
- [x] change logic for 
   - comment - could be set only after nhs_reviewed = true
   - [deactivate LE](https://uaehealthapi.docs.apiary.io/#reference/public.-medical-service-provider-integration-layer/legal-entities/deactivate-legal-entity) - could be changed only after nhs_reviewed = true
   - [verify LE by NHS](https://uaehealthapi.docs.apiary.io/#reference/public.-medical-service-provider-integration-layer/legal-entities/verify-legal-entity-by-nhs) - could be changed only after nhs_reviewed = true
- [x] change get LE by nhs' add new fields 
   - comment
   - nhs_reviewed with filters

progsmile:
add `nhs_verified = false`

svetavedmed:
- [ ] add nhsReview to get LE list with ordering nhsReview_ASC, nhsReview_DESC [PR](https://github.com/edenlabllc/ehealth.web/pull/507/files)

svetavedmed:
- [ ] add nhsReview to get LE list with ordering nhsReview_ASC, nhsReview_DESC [PR](https://github.com/edenlabllc/ehealth.web/pull/507/files)

svetavedmed:
- [ ] add nhsReview to get LE list with ordering nhsReview_ASC, nhsReview_DESC [PR](https://github.com/edenlabllc/ehealth.web/pull/507/files)

svetavedmed:
- [ ] should be forbidden to set "nhsReviewed": false by NHS. Only on update legal_entity "nhsReviewed" flag should (and it is) set to false.

PR #3769, opened 2018-12-05
Validate status and nhs_review on legal_entity mutations

progsmile:
#3469


apps/core/lib/core/legal_entities/legal_entities.ex
```diff
@@ -259,11 +260,19 @@ defmodule Core.LegalEntities do
     {:error, {:conflict, "LegalEntity is VERIFIED and cannot be VERIFIED."}}
   end
 
-  defp check_nhs_verify_transition(%LegalEntity{status: status}, _) when status != @status_active do
+  defp check_nhs_verify_transition(_, _), do: :ok
```
PavelVesnin: what for this verification when it has no other cases?

Issue #3407, opened 2018-11-05
GraphQL JSON error log


PavelVesnin:
Check that all application log is in JSON format

progsmile:
Found exception in json format on pod
`{"log_type":"error","message":"An exception was raised: ... }`

PR #3779, opened 2018-12-06
Log all errors as json from graphql

progsmile:
#3407

### DIFF
--- a/apps/graphql/lib/graphql_web/resolvers/helpers/errors.ex
+++ b/apps/graphql/lib/graphql_web/resolvers/helpers/errors.ex
@@ -1,10 +1,9 @@
 defmodule GraphQLWeb.Resolvers.Helpers.Errors do
   @moduledoc false
 
+  alias Core.Log
   alias Ecto.Changeset
   alias EView.Helpers.ChangesetValidationsParser
-
-  require Logger
 
   @unauthenticated_error_message "Unable to authenticate request"
   @forbidden_error_message "Current client is not allowed to access this resource"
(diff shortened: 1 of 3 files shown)


apps/graphql/lib/graphql_web/schema/capitation_contract_request_types.ex
```diff
@@ -27,7 +27,7 @@ defmodule GraphQLWeb.Schema.CapitationContractRequestTypes do
 
       # TODO: Replace it with `GraphQLWeb.Middleware.Filtering`
       middleware(GraphQLWeb.Middleware.FilterArgument)
-      resolve(&CapitationContractRequestResolver.list_contract_requests/2)
+      resolve(safe(&CapitationContractRequestResolver.list_contract_requests/2))
```
kiurchv: It's better to use middleware for this. I'm not sure that we should catch exceptions on the resolver level

progsmile: fully agree that's not clear solution, but for production it would be better to handle errors :)

apps/graphql/lib/graphql_web/resolvers/helpers/errors.ex
```diff
@@ -146,4 +145,20 @@ defmodule GraphQLWeb.Resolvers.Helpers.Errors do
...
+  def safe(function) when is_function(function, 3) do
+    fn parent, args, resolution -> do_apply(function, [parent, args, resolution]) end
+  end
+
+  defp do_apply(function, params) do
+    apply(function, params)
+  rescue
+    error ->
+      Log.error(%{"message" => "An exception was raised: #{inspect(error)}"})
+      {:error, Exception.message(error)}
```
progsmile:
```suggestion
      {:error, "Something went wrong"}
```

PR #3784, opened 2018-12-06
Add reimbursement contract request schema

kiurchv:


apps/core/lib/elixir/calendar/date_interval.ex
```diff
@@ -39,6 +39,18 @@ defmodule Date.Interval do
   defp raw_from_edtf(<<?., ?., ?/, last::binary-size(10)>>), do: {:ok, nil, last}
   defp raw_from_edtf(_), do: {:error, :invalid_format}
 
+  @spec from_edtf!(String.t()) :: t
+  def from_edtf!(string) do
```
progsmile: How about removing this function? I've only added error handles for graphql resolvers

apps/graphql/features/contexts/context.exs
```diff
@@ -79,6 +92,31 @@ defmodule GraphQL.Features.Context do
...
+      {:ok, state}
+    end
+  )
+
+  given_(
+    ~r/^the following employees exist:$/,
+    fn state, %{table_data: table_data} ->
+      for row <- table_data do
+        attrs = prepare_attrs(Employee, row)
+        # FIXME: There are should be better way to set party_id foreign key for employee
```
progsmile: How about use only TODO?

apps/graphql/lib/graphql_web/loaders/il.ex
```diff
@@ -3,7 +3,7 @@ defmodule GraphQLWeb.Loaders.IL do
 
   import Ecto.Query, only: [where: 2]
 
-  alias Core.ContractRequests.CapitationContractRequest
+  alias Core.ContractRequests.{CapitationContractRequest, ReimbursementContractRequest}
```
progsmile: How about not doing so? Most codebase uses full aliases, and it's not convenient to read from single line

progsmile: And refactoring opportunities are really poor

kiurchv: I'd argue on this

apps/graphql/features/contexts/context.exs
kiurchv: Nope. Other comment tags also useful

apps/core/lib/elixir/calendar/date_interval.ex
kiurchv: For what? It's a good tone to provide alternate ways for decoding values, in a similar way this done in the Elixir's `Date` and `DateTime` modules.

progsmile: I meant using `from_edtf` instead of `from_edtf!` which throws. Our current approach is not `let is crash` especially on parsing date interval
So it can be present, but not for using in graphql

PR #3833, opened 2018-12-11
add root Query fields for MedicalProgram

kiurchv:


apps/graphql/features/contexts/context.exs
```diff
@@ -207,6 +217,33 @@ defmodule GraphQL.Features.Context do
...
+          medicalPrograms(first: $first) {
+            nodes {
+              id
+              databaseId
+            }
+          }
+        }
+      """
+
+      variables = %{first: Jason.decode!(count)}
```
progsmile: `Jason.decode!(count)` suppose it's incorrect

apps/graphql/features/contexts/context.exs
```diff
@@ -305,10 +342,41 @@ defmodule GraphQL.Features.Context do
...
+          medicalPrograms(first: $first, filter: $filter) {
+            nodes {
+              #{field}
+            }
+          }
+        }
+      """
+
+      variables = %{
+        first: Jason.decode!(count),
```
progsmile: why count is required to decode?

PavelVesnin: same question. Why you decided to use Jason.decode! here?

apps/graphql/features/contexts/context.exs
kiurchv: All table values are treated as JSON since #3784. I'll replace all other conversions in the next commit

PR #3845, opened 2018-12-12
Return only certain type of contracts and contractRequests on graphql

progsmile:
#3844


apps/graphql/test/web/resolvers/capitation_contract_request_resolver_test.exs
```diff
@@ -152,6 +152,24 @@ defmodule GraphQLWeb.CapidationContractRequestResolverTest do
   end
 
   describe "list" do
```
kiurchv: It would be great if we'll start writing feature based tests for simple cases like this one and will use ExUnit tests as a backup solution — only for complex cases which is hard to cover with Gherkin.

PR #3848, opened 2018-12-12
add filter on party's full name to employees field

kiurchv:


apps/graphql/features/contexts/context.exs
```diff
@@ -90,6 +90,16 @@ defmodule GraphQL.Features.Context do
     end
   )
 
+  given_(
+    ~r/^there are (?<count>\d+) employees exist$/,
+    fn state, %{count: count} ->
+      count = String.to_integer(count)
```
PavelVesnin: why not `Jason.decode!()` ?

kiurchv: 😱

kiurchv: Fixed

Issue #3366, opened 2018-10-29
Change DS validation for merged LE


danakonovalenko:
Merge must be signed by employee with DS as NHS employee.
- [x] [design](https://edenlab.atlassian.net/wiki/x/zQDiJQ)
- [x] implement
- [ ] test

warehousman:
@danakonovalenko still open?

danakonovalenko:
this task is not done yet

PR #3856, opened 2018-12-13
Add legal_entity edrpou check on merge_legal_entities job

progsmile:
#3366


apps/core/test/support/prm_factories/party_factory.ex
```diff
@@ -12,7 +12,7 @@ defmodule Core.PRMFactories.PartyFactory do
           second_name: "Миколайович",
           birth_date: ~D[1991-08-19],
           gender: "MALE",
-          tax_id: sequence("222222222"),
+          tax_id: random_tax_id(),
```
PavelVesnin: why not sequence? Random can collide, when tax_id must be unique

apps/graphql/test/web/resolvers/legal_entity_merge_job_resolver_test.exs
```diff
@@ -90,6 +91,26 @@ defmodule GraphQLWeb.LegalEntityMergeJobResolverTest do
...
+
+      resp =
+        conn
+        |> put_client_id(UUID.generate())
+        |> post_query(@query, input_signed_content(signed_content))
+        |> json_response(200)
+
+      resp_entity = get_in(resp, ~w(data legalEntityMergeJobs))
+
+      refute resp_entity
```
PavelVesnin: could be in one line

apps/core/test/support/prm_factories/party_factory.ex
progsmile: had issues with unique_constraint

progsmile: I've check it again

progsmile: done with sequence under the hood

Issue #3599, opened 2018-11-20
Sign contract_request by NHS graphql


IrinaRevutska:
[spec](https://edenlab.atlassian.net/wiki/spaces/EH/pages/524320940/Sign+Contract+Request+by+NHS)
p.4 - capitation only
p.8 - new
[DataModel](https://edenlab.atlassian.net/wiki/spaces/EH/pages/671252546/Contract+Reimbursment+version+Data+model)

PR #3857, opened 2018-12-13
sign ContractRequest by NHS

PavelVesnin:
#3599


apps/core/lib/core/contract_requests/contract_requests.ex
```diff
@@ -394,20 +400,22 @@ defmodule Core.ContractRequests do
...
            ),
          update_params <-
-           params
-           |> Map.put("updated_by", user_id)
-           |> Map.put("status", CapitationContractRequest.status(:nhs_signed))
-           |> Map.put("nhs_signed_date", Date.utc_today())
-           |> Map.put("printout_content", printout_content),
+           Map.merge(pack.request_params, %{
+             "updated_by" => user_id,
+             "status" => CapitationContractRequest.status(:nhs_signed),
```
progsmile: How about extracting status to attribute? so there is no mention about particular contract type (in our case `CapitationContractRequest `)

Issue #3524, opened 2018-11-13
Assign reimbursement contract request graphql


IrinaRevutska:
[DataModel](https://edenlab.atlassian.net/wiki/spaces/EH/pages/671252546/Contract+Reimbursment+version+Data+model)

PR #3859, opened 2018-12-13
Reimbursement Contract Request - update assignee

PavelVesnin:
#3524


apps/core/lib/core/contract_requests/contract_requests.ex
```diff
@@ -39,6 +39,7 @@ defmodule Core.ContractRequests do
   @capitation CapitationContractRequest.type()
   @reimbursement ReimbursementContractRequest.type()
 
+  @new CapitationContractRequest.status(:new)
```
kiurchv: Consider to move statuses into `ContractRequest` module

PR #3871, opened 2018-12-17
Handle graphql exceptions with error view

progsmile:
#3407


apps/graphql/lib/graphql_web/resolvers/helpers/error_view.ex
```diff
@@ -0,0 +1,16 @@
...
+  alias Core.Log
+
+  @internal_error_templates ["500.json", "501.json", "503.json", "505.json"]
+
+  def render(template, %{reason: error}) when template in @internal_error_templates do
+    Log.error(%{"message" => "An exception was raised: #{inspect(error)}"})
+
+    %{
+      message: "Something went wrong",
+      extensions: %{code: "INTERNAL_SERVER_ERROR"}
```
kiurchv: Maybe it's worth to include request ID here?

progsmile: Exists in headers as `x-request-id`

kiurchv: Ok, header existence enough for us

PR #3911, opened 2018-12-20
Replace FilterArgument middleware with Filtering

progsmile:
#3873


apps/graphql/lib/graphql_web/loaders/prm.ex
```diff
@@ -39,9 +39,5 @@ defmodule GraphQLWeb.Loaders.PRM do
 
   def query(queryable, _), do: queryable
 
-  defp filter(query, [{:merged_from_legal_entity, filter} | tail]) do
-    filter(query, [{:merged_from, filter} | tail])
-  end
-
-  defp filter(query, filter), do: Search.filter(query, filter)
+  defp filter(query, filter), do: Filtering.filter(query, filter)
```
PavelVesnin: why not to import `Filtering.filter/2`?

apps/graphql/lib/graphql_web/schema/division_types.ex
```diff
@@ -48,7 +48,7 @@ defmodule GraphQLWeb.Schema.DivisionTypes do
 
   input_object :division_filter do
     field(:database_id, :id)
-    field(:name, :string_like)
+    field(:name, :string)
```
PavelVesnin: 👍

apps/graphql/lib/graphql_web/loaders/prm.ex
progsmile: Remember what are cases to import?

PavelVesnin: No. What cases?

progsmile: Import infrastructure functions, aliasing business, so here should be import (but personally I would aliasing everything)

progsmile: oh, ok, I've got the point

Issue #3036, opened 2018-08-30
Get Person Details NHS admin v2


danakonovalenko:
As NHS admin I want to view person details, so we need to implement service according to  [schema](https://github.com/edenlabllc/ehealth.schema/blob/master/__schema__/persons.graphql) / [features](https://github.com/edenlabllc/ehealth.web/blob/master/packages/admin/e2e/__features__/person_details.feature)

lymychp:
documents should be a parameter of declaration, not person

PR #3984, opened 2019-01-09
Connect graphql to rpc cluster

progsmile:
#3036


apps/graphql/lib/application.ex
```diff
@@ -12,6 +12,16 @@ defmodule GraphQL.Application do
       supervisor(GraphQLWeb.Endpoint, [])
     ]
 
+    children =
+      if Application.get_env(:graphql, :env) == :prod do
```
AlexKovalevych: are you sure you have this config variable?

progsmile: no

PR #3987, opened 2019-01-09
EHealth-3358: Reimbursement CR without attached documents

PavelVesnin:


apps/core/lib/core/contract_requests/storage.ex
```diff
@@ -3,31 +3,37 @@ defmodule Core.ContractRequests.Storage do
...
 
     with {:ok, %{"data" => %{"secret_url" => statute_url}}} <-
-           @media_storage_api.create_signed_url(
-             "PUT",
-             get_bucket(),
-             "media/upload_contract_request_statute.pdf",
-             id,
-             []
-           ),
+           @media_storage_api.create_signed_url("PUT", get_bucket(), @statute.upload_path, id, []),
```
progsmile: Looks like low level api, maybe will refactor it in future

apps/core/lib/core/contract_requests/storage.ex
```diff
@@ -10,6 +10,8 @@ defmodule Core.ContractRequests.Storage do
   @signature_api Application.get_env(:core, :api_resolvers)[:digital_signature]
   @media_storage_api Application.get_env(:core, :api_resolvers)[:media_storage]
 
+  @reimbursement "REIMBURSEMENT"
```
progsmile: Better to grab that from ReimbursementContractRequest?

Issue #3035, opened 2018-08-30
Search for a Person NHS admin v2


danakonovalenko:
As NHS admin I want to be able to search for a person using next possible filter parameters:

---
phone_number
birth_date

---
tax_id
passport_number

---
at least one of field in each block must be required

and sort by birth_date, tax_id, national_id

We need to implement WS according to [spec](https://github.com/edenlabllc/ehealth.schema/blob/master/__schema__/persons.graphql)

lymychp:
@warehousman @danakonovalenko please review list of combinations

danakonovalenko:
[test cases](https://edenlab.atlassian.net/wiki/x/WIArKg)
- [ ] street_type and zip should be optional
- [ ] there is no declarations connection yet

PR #4033, opened 2019-01-17
Return nil on rpc loader when does not find value under batch key

progsmile:
#3035


apps/graphql/lib/graphql_web/schema.ex
```diff
@@ -95,7 +95,7 @@ defmodule GraphQLWeb.Schema do
 
   def context(ctx) do
     loader =
-      Dataloader.new()
+      Dataloader.new(get_policy: :return_nil_on_error)
```
PavelVesnin: 👍

PR #4046, opened 2019-01-18
set globalId for LegalEntityDeactivationJob

PavelVesnin:


apps/graphql/lib/graphql_web/schema/legal_entity_merge_job_types.ex
```diff
@@ -73,6 +73,7 @@ defmodule GraphQLWeb.Schema.LegalEntityMergeJobTypes do
   end
 
   node object(:legal_entity_merge_job) do
+    field(:id, non_null(:id))
```
progsmile: It should work without this field

progsmile: Does it?

PavelVesnin: you're right, mistake

PR #4060, opened 2019-01-23
Add declaration approve/reject mutations

kiurchv:


apps/graphql/lib/graphql_web/resolvers/declaration_resolver.ex
```diff
@@ -55,18 +56,46 @@ defmodule GraphQLWeb.Resolvers.DeclarationResolver do
...
+      declaration = api_response_to_ecto_struct(Declaration, declaration)
 
       {:ok, %{declaration: declaration}}
     else
       err -> render_error(err)
     end
   end
+
+  def terminate_declaration(
+        %{id: id, reason_description: reason_description},
```
progsmile: reason_description is optional, do not matcch

apps/graphql/lib/graphql_web/resolvers/declaration_resolver.ex
```diff
@@ -55,18 +56,46 @@ defmodule GraphQLWeb.Resolvers.DeclarationResolver do
...
+    update_declaration(id, patch, headers)
+  end
+
+  def reject_declaration(%{id: id}, %{context: %{consumer_id: consumer_id, headers: headers}}) do
+    patch = %{"status" => @status_rejected, "updated_by" => consumer_id}
+
+    update_declaration(id, patch, headers)
+  end
+
+  defp update_declaration(id, patch, headers) do
```
progsmile: how about `update_data`, or `params` instead of `patch`?

apps/graphql/lib/graphql_web/schema/declaration_types.ex
```diff
@@ -42,9 +42,43 @@ defmodule GraphQLWeb.Schema.DeclarationTypes do
...
+        field(:declaration, :declaration)
+      end
+
+      middleware(ParseIDs, id: :declaration)
+      resolve(&DeclarationResolver.reject_declaration/2)
+    end
+
     payload field(:terminate_declaration) do
       meta(:scope, ~w(declaration:terminate))
-      meta(:client_metadata, ~w(client_id client_type)a)
```
progsmile: client_id is required to be in headers, would it work?

apps/graphql/lib/graphql_web/resolvers/declaration_resolver.ex
kiurchv: But `patch` means the same as update data and it definitely not params, agree?

progsmile: ok, definitely not params

apps/graphql/lib/graphql_web/schema/declaration_types.ex
kiurchv: We aren't authorizing request by client_id here, only client type authorization is needed

Issue #3925, opened 2018-12-21
Terminate Declaration mutation


danakonovalenko:
Declaration in status `active` can be terminated. In order to do that we should implement mutation
```
 terminateDeclaration(
    input: TerminateDeclarationInput!
  ): TerminateDeclarationPayload
```
[schema](https://github.com/edenlabllc/ehealth.schema/blob/master/__schema__/index.graphql)
[features](https://github.com/edenlabllc/ehealth.web/blob/master/packages/admin/e2e/__features__/declaration_details.feature)

lymychp:
blocks edenlabllc/ehealth.web#636

danakonovalenko:
I can't complete happy pass as I receive an error - `Current client is not allowed to access this resource`,
however token belongs to NHS
[test cases](https://edenlab.atlassian.net/wiki/x/CgBtKg)

danakonovalenko:
status_reason should be changed from `varchar(255)` to `text` as for now we receive 500 error
```
 "extensions": {
        "code": "BAD_REQUEST",
        "exception": "%{\"error\" => %{\"errors\" => %{\"detail\" => \"Internal server error\"}}, \"meta\" => %{\"code\" => 500, \"request_id\" => \"86aceff7-7039-41d5-b6b4-161fdb77f360#20467\", \"type\" => \"object\", \"url\" => \"http://graphql-svc.il/declarations/66d5524c-a52c-4690-a7ac-f3a632b2213c/actions/terminate\"}}"
      }
```

PR #4070, opened 2019-01-23
Remove user check on terminate declaration process on graphql

progsmile:
#3925


apps/core/lib/core/declarations/api.ex
```diff
@@ -201,16 +201,25 @@ defmodule Core.Declarations.API do
...
     else
       {:status, false} -> {:conflict, "Declaration is not active"}
-      {:person_id, false} -> {:error, :forbidden}
       error -> error
     end
   end
 
+  defp do_check_user_access(true, _, _), do: :ok
+
+  defp do_check_user_access(false, user, declaration) do
```
kiurchv: Arent we can match this by pattern, can we?
```suggestion
  defp do_check_user_access(false, %{"person_id" => id}, %{"person" => %{"id" => id}}), do: :ok

    defp do_check_user_access(false, _, _), do: {:error, :forbidden}
```

Issue #4081, opened 2019-01-24
500 on reject medication dispense


svetavedmed:
```{"error":{"errors":{"detail":"Internal server error"}},"meta":{"code":500,"request_id":"d1b4ae6d-23d6-4cb5-ad7d-d10e0b521532#123360","type":"object","url":"http://api-svc.il/api/medication_dispenses/8f5819cb-da60-4302-82ce-52e3af83ca44/actions/reject"}}```

AlexKovalevych:
@svetavedmed  which environment?

svetavedmed:
demo

PR #4098, opened 2019-01-29
fix references load on reject medication_dispense

imax-iva:


imax-iva:
https://github.com/edenlabllc/ehealth.api/issues/4081

Issue #4092, opened 2019-01-28
Refactor declaration mutations to use rpc instead of rest


progsmile:

progsmile:
Please check next actions with declaration on rest and graphql:
- terminate
- approve / reject

PR #4100, opened 2019-01-29
Refactor declaration update and terminate functions using rpc

progsmile:
#4092


apps/core/lib/core/declarations/api.ex
```diff
@@ -308,12 +330,21 @@ defmodule Core.Declarations.API do
     end
   end
 
+  def do_update_declaration(id, params) do
```
AlexKovalevych: why do you need this function at all?

apps/core/lib/core/declarations/api.ex
```diff
@@ -308,12 +330,21 @@ defmodule Core.Declarations.API do
...
+    with {:ok, declarations} <- @rpc_worker.run("ops", Core.Rpc, :update_declaration, [id, params]) do
+      {:ok, declarations}
+    end
+  end
+
+  def do_terminate_declaration(id, %{} = params) do
```
AlexKovalevych: same question

apps/core/lib/core/jason_protocols.ex
```diff
@@ -17,3 +17,11 @@ defimpl Jason.Encoder, for: Core.MedicationRequestRequest.EmbeddedData do
     Encode.map(Map.take(value, Core.MedicationRequestRequest.EmbeddedData.__schema__(:fields)), opts)
   end
 end
+
+defimpl Jason.Encoder, for: Core.Declarations.Declaration do
```
AlexKovalevych: why can't you render declarations at the ops service and receive a map with string keys in response? Cause as i understand, there is no module `Core.Declarations.Declaration` in the ehealth.api

apps/graphql/test/web/resolvers/declaration_resolver_test.exs
```diff
@@ -416,6 +393,14 @@ defmodule GraphQLWeb.DeclarationResolverTest do
          }}
       end)
 
+      expect(RPCWorkerMock, :run, fn _, _, :get_declaration, _ -> declaration end)
+      expect(RPCWorkerMock, :run, fn _, _, :get_person_by_id, _ -> {:ok, person} end)
```
AlexKovalevych: hm, how will it work, if the `:run` function is the same?

apps/core/lib/core/declarations/api.ex
progsmile: ok

apps/core/lib/core/declarations/api.ex
progsmile: ok

apps/graphql/test/web/resolvers/declaration_resolver_test.exs
progsmile: by order of mocks

apps/core/lib/core/jason_protocols.ex
progsmile: Previously there was proxy response.. and we received declaration as string-keyed map. Now it's rpc call with structure, so may remove it in view

Issue #4035, opened 2019-01-17
Get my merge requests list


danakonovalenko:
As nhs admin reviewer I want to be able to get my merge requests in statuses ('POSTPONE', 'NEW')

In order to do that we need to implement WS based on [logic](https://edenlab.atlassian.net/wiki/x/nQJUKg) and schema

- [x] implement
- [ ] test

ephor:
Blocked by edenlabllc/ehealth.api#4038

danakonovalenko:
there is a problem with checking role, I still get merge requests list/id without required role `NHS REVIEWER` in case I have required scope

svetavedmed:
- [ ] in case i do not have merge requests in status NEW i still receive 'canAssignNew: false'
![Screen Shot 2019-02-11 at 9.29.49 AM.png](https://zube.io/files/edenlabllc/f990b54157a7814425abca6107a967b9-screen-shot-2019-02-11-at-9-29-49-am.png)

progsmile:
Now, max requests count is 5

danakonovalenko:
in config `MAX_POSTPONED_MANUAL_MERGE_REQUESTS`= 5, however now it's I can have more. this parameters doesn't work

progsmile:
This is not limit for query
It's particular count which should be checked on assign

PR #4130, opened 2019-02-05
feat: implement merge_request list and get_by_id

progsmile:
#4035

### DIFF
--- a/apps/core/lib/core/medication_requests/renderer.ex
+++ b/apps/core/lib/core/medication_requests/renderer.ex
@@ -6,12 +6,11 @@ defmodule Core.MedicationRequests.Renderer do
   alias Core.Employees.Employee
   alias Core.LegalEntities.Renderer, as: LegalEntitiesRenderer
   alias Core.MedicalPrograms.MedicalProgram
-  alias Core.MedicationRequestRequest.Renderer, as: MedicationRequestRequestRenderer
+  alias Core.Persons.Renderer, as: PersonsRenderer
 
   def render("show.json", medication_request) do
     legal_entity = medication_request["legal_entity"]
     created_at = Timex.parse!(medication_request["created_at"], "{YYYY}-{0M}-{D}")
-    person = MedicationRequestRequestRenderer.render_person(medication_request["person"], created_at)
 
     response =
       medication_request
@@ -39,7 +38,7 @@ defmodule Core.MedicationRequests.Renderer do
         "employee" => render("employee.json", medication_request["employee"]),
         "division" => render("division.json", medication_request["division"]),
         "medication_info" => render("medication_info.json", medication_request),
-        "person" => person
+        "person" => render_person(medication_request["person"], created_at)
       })
 
     if Map.get(medication_request, "medical_program") do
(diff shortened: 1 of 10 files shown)


apps/graphql/test/unit/middleware/check_user_role_test.exs
```diff
@@ -0,0 +1,94 @@
...
+
+      field :forbidden, list_of(:resource) do
+        middleware(&put_client_id/2)
+        middleware(CheckUserRole, role: "ADMIN")
+      end
+
+      field :error, list_of(:resource) do
+        middleware(&put_client_id/2)
+        middleware(CheckUserRole, role: "ADMIN")
+      end
```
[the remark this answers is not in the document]
PavelVesnin: I don't see any difference in field definition except of naming.
Why you can't use one field instead of two? E.g. `:invalid`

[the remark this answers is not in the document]
progsmile: https://github.com/edenlabllc/ehealth.api/pull/4130/files/9f212b1e270708a2eea8138d022ccf33647af5c5#diff-5d777dc6942b69ae634ee99d5b1acf08R86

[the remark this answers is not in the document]
progsmile: Error from MithrilMock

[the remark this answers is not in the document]
progsmile: Got it)

apps/graphql/lib/graphql_web/schema/merge_request_types.ex
```diff
@@ -0,0 +1,95 @@
...
+
+  object :merge_request_queries do
+    @desc "Get pending MergeRequests"
+    connection field(:pending_merge_requests, node_type: :merge_request) do
+      meta(:scope, ~w(merge_request:read))
+      meta(:client_metadata, ~w(client_id client_type consumer_id)a)
+      meta(:allowed_clients, ~w(NHS))
+      arg(:order_by, :merge_request_order_by)
+
+      middleware(CheckUserRole, role: "NHS REVIEWER")
```
kiurchv: Maybe we should depend on meta attribute like we do for scope- and client-based authorization?

apps/graphql/test/web/resolvers/merge_request_resolver_test.exs
```diff
@@ -0,0 +1,208 @@
...
+        |> put_client_id()
+        |> put_consumer_id()
+        |> put_scope("merge_request:read")
+        |> post_query(@merge_request_by_id_query, variables)
+        |> json_response(200)
+
+      %{"errors" => [error]} = resp_body
+
+      refute get_in(resp_body, ~w(data mergeRequest))
+      assert "NOT_FOUND" == error["extensions"]["code"]
```
kiurchv: I believe that we should return `null` without any errors when we can't get merge request by some reason

apps/graphql/lib/graphql_web/schema/merge_request_types.ex
progsmile: Firstly I've implemented it with option, but than decided not to over-engineering and use only `client_id`
There begins mapping from `user_id` -> `consumer_id`, and 1 more option `user_ids`. So that's would be done only when it's used

apps/graphql/test/web/resolvers/merge_request_resolver_test.exs
progsmile: Suppose that's common behaviour around application to return such error 
If so, we should return null from error_renderer

kiurchv: We should receive `nil` from the DB when we cannot find merge request with given ID

progsmile: But it would be inconsistent with other not found responses

PR #4150, opened 2019-02-07
feat(graphql): add `email`, `emergencyContact` and `confidantPersons` fields to `Person` type

kiurchv:


apps/core/test/support/mpi_factories/person_factory.ex
```diff
@@ -94,6 +94,49 @@ defmodule Core.MPIFactories.PersonFactory do
...
+      end
+
+      def confidant_person_factory do
+        %{
+          relation_type: Enum.random(["PRIMARY", "SECONDARY"]),
+          first_name: random_first_name(),
+          last_name: random_last_name(),
+          second_name: random_second_name(),
+          birth_date: "1996-12-12",
+          birth_country: random_bith_country(),
```
PavelVesnin:
```suggestion
          birth_country: random_birth_country(),
```

apps/graphql/test/web/resolvers/person_resolver_test.exs
```diff
@@ -186,7 +227,7 @@ defmodule GraphQLWeb.PersonResolverTest do
       assert person.id == resp_entity["databaseId"]
 
       assert Enum.all?(
-               ~w(firstName lastName secondName birthDate gender status birthCountry birthSettlement taxId unzr preferredWayCommunication insertedAt documents addresses phones authenticationMethods),
+               ~w(firstName lastName secondName birthDate gender status birthCountry birthSettlement taxId unzr preferredWayCommunication insertedAt documents addresses phones authenticationMethods emergencyContact confidantPersons),
```
progsmile: Suppose that this check is redundant: particular fields are asked and there would be error in case they are not resolved. So  `refute resp_body["errors"]` covers this

apps/graphql/test/web/resolvers/person_resolver_test.exs
```diff
@@ -210,6 +251,31 @@ defmodule GraphQLWeb.PersonResolverTest do
       Enum.each(resp_entity["phones"], fn phone ->
         assert Enum.all?(~w(type number), &Map.has_key?(phone, &1))
       end)
+
+      assert Enum.all?(~w(firstName lastName secondName phones), &Map.has_key?(resp_entity["emergencyContact"], &1))
```
progsmile: here we may just `assert resp_entity["emergencyContact"]`

apps/graphql/test/web/resolvers/person_resolver_test.exs
kiurchv: Ok, so should we delete all these assertions?

apps/graphql/lib/graphql_web/schema/person_types.ex
```diff
@@ -174,6 +178,45 @@ defmodule GraphQLWeb.Schema.PersonTypes do
...
+    field(:phones, non_null(list_of(:phone)))
+  end
+
+  object :confidant_person do
+    field(:relation_type, non_null(:confidant_person_relation_type))
+    field(:first_name, non_null(:string))
+    field(:last_name, non_null(:string))
+    field(:second_name, :string)
+    # TODO: This field should return :date type
+    field(:birth_date, :string)
```
progsmile: - may cast on mpi, but it's used on rest
- or in resolver 
What you think?

kiurchv: This issue would be resolved if we ever define embedded schemas for person's JSON fields on MPI. I'd left it as is for now

apps/graphql/test/web/resolvers/person_resolver_test.exs
progsmile: For now let it be as it is, they are redundant, but iron )

progsmile: You may delete for first level fields

Issue #4112, opened 2019-01-31
add death_date to registers with type "DEATH_REGISTRATION"


svetavedmed:
- change .csv file, add death_date. Example of a new file [per.csv](https://zube.io/files/edenlabllc/d131b4ee0770ce05382251cbd9a12ec9-per.csv)
- death_date is optional for csv file
- add MPI_ID to document type
- change validations [spec](https://edenlab.atlassian.net/wiki/x/ioCBF)
apiary
```
Validate that $document.type 
one of from dictionary (name='DOCUMENT_TYPE') or is MPI_ID. If type is not as in dictionary or not MPI_ID show the row in the errors field.
Validate MPI_ID
   in case MPI_ID is not UUID ("^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$")
      show the row with not found MPI_ID with status "error"
   in case MPI_ID was not found in mpi.persons.id
      show the row with MPI_ID with status "not_found"
   in case MPI_ID was found, but could not be deactivated
      show row with MPI_ID with status "processed"
Validate death_date if exists
   in case death_date <> valid date ()
      show the row with MPI_ID with status "date_error"
   in case death_date does not exist
      show the row with MPI_ID with status "date_error"
   in case death_date year < 1900
      show the row with MPI_ID with status "date_error"
Validate that $type one of from dictionary (name='REGISTER_TYPE'). If type is not as in dictionary show the 422 error 'Incorrect register type'

Validate headers in file: type, number, death_date. In case of error show the 422 error 'Incorrect headers in file' 
```
If the person was matched then
add update mpi.persons.death_date = $death_date

PR #4155, opened 2019-02-08
feat: add death_date to registers; add MPI_ID type

progsmile:
#4112


apps/core/lib/core/registers/api.ex
```diff
@@ -212,12 +214,14 @@ defmodule Core.Registers.API do
...
+  defp validate_csv_headers(headers) do
+    values = headers |> Map.keys() |> Enum.map(&String.downcase/1) |> MapSet.new()
 
-  defp valid_csv_headers?(_) do
-    Error.dump("Invalid CSV headers")
+    cond do
+      values == MapSet.new(["type", "number"]) -> :ok
+      values == MapSet.new(["type", "number", "death_date"]) -> :ok
+      true -> Error.dump("Invalid CSV headers")
+    end
```
PavelVesnin: Why you don't use pattern matching for this case?
FYI pattern matching is a little bit faster than `cond` and other control flow structures.

apps/core/lib/core/validators/dealth_date.ex
```diff
@@ -0,0 +1,14 @@
...
+  def validate(death_date) when is_binary(death_date) do
+    with {:ok, date} <- Date.from_iso8601(death_date),
+         true <- Timex.between?(date, valid_from_date(), Date.utc_today(), inclusive: true) do
+      :ok
+    else
+      _ -> :error
+    end
+  end
+
+  defp valid_from_date, do: ~D[1990-01-01]
```
PavelVesnin: is it business requirement?

PavelVesnin: date is wrong, from spec: in case death_date year < 1900

PavelVesnin: date is wrong, from spec: in case death_date year < 1900

apps/core/lib/core/registers/api.ex
progsmile: Because can match on other keys in headers

apps/core/lib/core/validators/dealth_date.ex
progsmile: thanks

apps/core/lib/core/validators/dealth_date.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Core.Validators.DeathDate do
+  @moduledoc false
+
+  def validate(death_date) when is_binary(death_date) do
+    with {:ok, date} <- Date.from_iso8601(death_date),
+         true <- Timex.between?(date, valid_from_date(), Date.utc_today(), inclusive: true) do
```
kiurchv:
```suggestion
         true <- Enum.member?(Date.range(date, valid_from_date()), Date.utc_today()) do
```

apps/core/lib/core/registers/api.ex
PavelVesnin: do we need exactly two or three headers?
E.g.
```
defp valid_csv_headers?(%{"type" => _, "number" => _}), do: :ok
defp valid_csv_headers?(%{"type" => _, "number" => _, "death_date" => _}), do: :ok
defp valid_csv_headers?(_), do: Error.dump("Invalid CSV headers")
```

progsmile: Input example `%{"type" => _, "number" => _, "lol" => _}`

PR #4169, opened 2019-02-11
fix(graphql): add consumer_id to check user role

progsmile:
#4035


apps/graphql/test/unit/middleware/check_user_role_test.exs
```diff
@@ -23,7 +23,7 @@ defmodule GraphQL.Unit.CheckUserRoleTest do
 
     query do
       field :resource_field, list_of(:resource) do
-        middleware(&put_client_id/2)
+        middleware(&put_auth_data/2)
```
kiurchv: I believe that this should happen in headers

Issue #3355, opened 2018-10-26
UA Addresses


danakonovalenko:
- [x] design dictionary
- [x] implement
- [x] test

blocked by edenlabllc/ehealth.api#3356

lymychp:
using bad rpc approach

PR #4251, opened 2019-02-19
feat(graphql): add `Settlement`, `District` and `Region` types

kiurchv:
connects #3355


apps/graphql/lib/graphql_web/resolvers/settlement_resolver.ex
```diff
@@ -0,0 +1,19 @@
+defmodule GraphQLWeb.Resolvers.SettlementResolver do
+  @moduledoc false
+
+  import GraphQLWeb.Resolvers.Helpers.Load, only: [response_to_ecto_struct: 2]
+
+  alias Core.Uaddresses
+  alias Core.Uaddresses.Settlement
+  alias Absinthe.Relay.Connection
```
progsmile: alias

Issue #4259, opened 2019-02-20
GraphQL Exception for invalid path to file in GCS


progsmile:
```
Request: POST /graphql
** (exit) an exception was raised:
[...]

PR #4261, opened 2019-02-20
Fix exception for invalid path to file in gcs

progsmile:
#4259


apps/core/lib/core/api/media_storage.ex
```diff
@@ -73,7 +73,19 @@ defmodule Core.API.MediaStorage do
...
 
-  def get_signed_content(secret_url), do: HTTPoison.get(secret_url)
+  def get_signed_content(secret_url) do
+    case HTTPoison.get(secret_url) do
+      {:ok, %{status_code: 200}} = result ->
+        result
+
+      {:ok, %{body: body}} ->
+        Logger.error("Fail to get signed_content, details: #{inspect(body)}")
+        {:error, {:conflict, "Fail to get signed_content"}}
```
PavelVesnin: Failed

apps/core/lib/core/contract_requests/storage.ex
```diff
@@ -150,9 +150,11 @@ defmodule Core.ContractRequests.Storage do
 
   def move_uploaded_documents(%RequestPack{} = pack, headers) do
     Enum.reduce_while([@statute, @additional_document], :ok, fn doc, _ ->
-      case md5_request_param_exist?(pack, doc.request_param_key) do
-        true -> move_file(pack, doc.upload_path, doc.permanent_path, headers)
-        false -> {:cont, :ok}
+      with true <- md5_request_param_exist?(pack, doc.request_param_key),
```
AlexKovalevych: why do you need reduce_while if you always cont with `:ok`?

progsmile: nice catch! I'll add error

PR #4263, opened 2019-02-20
feat: mpi rpc  get persons by ids

marinakr:


marinakr:
locked by  https://github.com/edenlabllc/mpi.api/pull/232

PR #4272, opened 2019-02-22
feat(graphql): add medication types

progsmile:


apps/graphql/lib/graphql_web/schema/medication_types.ex
```diff
@@ -0,0 +1,88 @@
...
+    field(:inserted_at, non_null(:naive_datetime))
+    field(:updated_at, non_null(:naive_datetime))
+  end
+
+  object :manufacturer do
+    field(:name, non_null(:string))
+    field(:country, non_null(:string))
+  end
+
+  enum :medicationform do
```
PavelVesnin: this form is identical to `innm_dosage_form`. Let's reuse it

progsmile: Suppose to use `medication_form` because `medication` it's parent for `innm_dosage`

PR #4285, opened 2019-02-22
MSP+Pharmacy

marinakr:


apps/ehealth/lib/ehealth/web/views/v2/legal_entity_view.ex
```diff
@@ -0,0 +1,54 @@
...
+
+    legal_entity
+    |> Map.take(@fields)
+    |> Map.put(:medical_service_provider, render_one(msp, __MODULE__, "medical_service_provider.json"))
+  end
+
+  def render("legal_entity.json", %{legal_entity: legal_entity}), do: legal_entity
+
+  def render("medical_service_provider.json", %{legal_entity: msp}),
+    do: %{licenses: msp.licenses, accreditation: msp.accreditation}
```
AlexKovalevych: `Map.take(msp, ~w(licenses accreditation)a)`

apps/core/lib/core/validators/v2/kveds.ex
```diff
@@ -0,0 +1,65 @@
+defmodule Core.Validators.V2.KVEDs do
+  @moduledoc """
+  KVED codes validator
+  """
+
+  import Ecto.Changeset
+
+  alias Core.Dictionaries
+  alias Core.LegalEntities.LegalEntity
+  alias Core.Validators.KVEDs
```
AlexKovalevych: probably it's better to alias is with separate name to avoid confusing, since the current module name is also `KVEDs`

apps/core/priv/repo/migrations/20190222190600_add_kveds_dictionaries_msp.exs
```diff
@@ -0,0 +1,21 @@
+defmodule Core.Repo.Migrations.AddKvedsDictionaries do
+  use Ecto.Migration
+  alias Ecto.UUID
+
+  def change do
+    execute("""
+    INSERT INTO dictionaries
```
AlexKovalevych: should it be part of `dictionaries.json` ?

PR #4321, opened 2019-02-28
feat(graphql): add query fields for `Medication`

kiurchv:


apps/ehealth/test/web/controllers/medication_dispense_controller_test.exs
```diff
@@ -2924,6 +2924,7 @@ defmodule EHealth.Web.MedicationDispenseControllerTest do
       {:ok, %{conn: conn}}
     end
 
+    @tag :w
```
PavelVesnin: `:w`? you mean `:wip`?
we are using tag `pending `

apps/graphql/features/contexts/context.exs
```diff
@@ -247,6 +260,30 @@ defmodule GraphQL.Features.Context do
...
+        attrs = prepare_attrs(Medication.Ingredient, row)
+        insert(:prm, :ingredient_medication, attrs)
+      end
+
+      {:ok, state}
+    end
+  )
+
+  given_(
+    ~r/^the following innm dosages exist:$/,
```
PavelVesnin:
```suggestion
    ~r/^the following INNM dosages exist:$/,
```

apps/ehealth/test/web/controllers/medication_dispense_controller_test.exs
progsmile: just to mark that this test fails

Issue #4001, opened 2019-01-11
Error 500 when user tries to execute Get Contract Request Detail


svetavedmed:
Error 500 when user tries to execute Get Contract Request Detail with deactivated divisions

PR #4325, opened 2019-02-28
fix: render only active divisions on contract requests show endpoint

progsmile:
#4001


apps/ehealth/test/web/controllers/contract_request/capitation_controller_test.exs
```diff
@@ -1756,20 +1757,54 @@ defmodule EHealth.Web.ContractRequest.CapitationControllerTest do
...
       expect(MediaStorageMock, :create_signed_url, 2, fn _, _, id, resource_name, _ ->
         {:ok, %{"data" => %{"secret_url" => "http://url.com/#{id}/#{resource_name}"}}}
       end)
 
-      resp =
-        conn
-        |> put_client_id_header(context.legal_entity_id_1)
-        |> get(contract_request_path(conn, :show, @capitation, context.contract_request_id_1))
-        |> json_response(200)
+      assert resp =
```
PavelVesnin: you don't need assert here

PR #4401, opened 2019-03-05
set LE addresses.settlement_id value as GlobalID

PavelVesnin:


apps/graphql/lib/graphql_web/schema/legal_entity_types.ex
```diff
@@ -16,6 +16,8 @@ defmodule GraphQLWeb.Schema.LegalEntityTypes do
     connection field(:legal_entities, node_type: :legal_entity) do
       meta(:scope, ~w(legal_entity:read))
 
+      middleware(ParseIDs, filter: [addresses: [settlement_id: :settlement]])
```
kiurchv: Let's move it after args, but before Filtering middleware

PR #4425, opened 2019-03-07
feat: tsrange filtering

kiurchv:


apps/core/lib/core/ecto/date_range.ex
```diff
@@ -0,0 +1,150 @@
...
+          lower: boundary_t(),
+          upper: boundary_t(),
+          lower_inclusive: boolean(),
+          upper_inclusive: boolean()
+        }
+
+  @type boundary_t :: Date.t() | nil
+
+  @open_boundary ".."
+  @interval_designator "/"
```
progsmile: @interval_separator

apps/core/lib/core/ecto/date_range.ex
```diff
@@ -0,0 +1,150 @@
...
+
+  @spec new(boundary_t(), boundary_t(), Keyword.t()) :: {:ok, t} | {:error, atom}
```
progsmile: t(), atom()

apps/core/lib/core/ecto/date_range.ex
kiurchv: According to Wikipedia, it seems that in spec it named as "interval_designator"
https://en.wikipedia.org/wiki/ISO_8601#Time_intervals

apps/core/lib/core/ecto/date_range.ex
```diff
@@ -0,0 +1,150 @@
...
+      _ -> :ok
+    end
+  end
+
+  def check_boundaries(%Date{}, %Date{}), do: {:error, :calendars_mismatch}
+  def check_boundaries(nil, nil), do: {:error, :invalid_boundaries}
+  def check_boundaries(_, _), do: :ok
+
+  @spec from_iso8601(String.t()) :: {:ok, t} | {:error, atom}
+  def from_iso8601(string) do
```
progsmile: how about `value`, `input`
string looks like type

apps/core/lib/core/ecto/date_range.ex
kiurchv: Explain better

apps/core/lib/core/ecto/date_range.ex
```diff
@@ -0,0 +1,150 @@
...
+  @spec to_iso8601(t, :extended | :basic) :: String.t()
+  def to_iso8601(%__MODULE__{lower: lower, upper: upper}, format \\ :extended) do
+    boundary_to_string(lower, format) <> @interval_designator <> boundary_to_string(upper, format)
+  end
+
+  defp boundary_to_string(nil, _), do: @open_boundary
+  defp boundary_to_string(date, format), do: Date.to_iso8601(date, format)
+
+  @impl Ecto.Type
+  def type, do: :daterange
```
progsmile: move `@impls` it to top of the file

apps/core/lib/core/ecto/date_range.ex
kiurchv: I don't see anything bad here. Elixir's `Date` and `DateTime` use the same name for parsing functions

apps/core/lib/core/ecto/date_range.ex
kiurchv: I disagree with this. I've added them here to distinguish behavior implementations from rest of the module public API

apps/graphql/test/web/resolvers/reimbursement_contract_resolver_test.exs
```diff
@@ -58,129 +58,6 @@ defmodule GraphQLWeb.ReimbursementContractResolverTest do
   end
 
   describe "list" do
-    test "return all for NHS client", %{conn: conn} do
```
progsmile: Are you sure that there is no need of removed tests? Does bdd covers all?

kiurchv: Yep. Removed tests are already covered in Gherkin features

apps/core/lib/core/ecto/date_range.ex
progsmile: in spec, why `boundary_t` with bracers and `atom`, `t` without?

apps/core/lib/core/ecto/date_range.ex
progsmile: ok

apps/core/lib/core/ecto/date_range.ex
kiurchv: Because `atom` is built-in type and local `t` is a well-known type name and `boundary_t` — not

apps/core/lib/core/ecto/date_range.ex
progsmile: It still ok because they are separated, however module has behaviour, and in my view it should implement it's impls first

apps/graphql/lib/graphql_web/resolvers/employee_resolver.ex
```diff
@@ -1,11 +1,11 @@
 defmodule GraphQLWeb.Resolvers.EmployeeResolver do
   @moduledoc false
 
-  import Ecto.Query, only: [join: 4, where: 3, order_by: 2]
+  import Ecto.Query, only: [order_by: 2]
+  import GraphQL.Filters.Base, only: [filter: 2]
```
PavelVesnin: i believe is not related to PR

apps/core/lib/core/ecto/date_range.ex
progsmile: that's really primitive code style question, but we need to find consistent way to write type annotations. Samples:
`@spec new(boundary_t, boundary_t, Keyword.t) :: {:ok, t} | {:error, atom}` or
`@spec new(boundary_t(), boundary_t(), Keyword.t()) :: {:ok, t()} | {:error, atom()}`
In my view first is better - no need to write `()`

apps/core/lib/core/ecto/timestamp_range.ex
```diff
@@ -0,0 +1,195 @@
+defmodule Core.Ecto.TimestampRange do
```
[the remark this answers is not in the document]
kiurchv: I have no idea how to do this now. Since these modules represent different DB types I don't think that this is possible

apps/graphql/lib/graphql_web/resolvers/employee_resolver.ex
kiurchv: I don't see any sense in adopting legacy filter functionality for new value types, so I've migrated all GraphQL codebase to ecto_filter

apps/core/lib/core/ecto/date_range.ex
progsmile: @PavelVesnin @kiurchv what do you think?

apps/graphql/lib/graphql_web/resolvers/employee_resolver.ex
PavelVesnin: agree. So first PR should change Filtering to Filters.Base,
next implement DateRange

kiurchv: Actually, I'm planning to merge all this PR as an individual commits

PavelVesnin: That's good, but it's not obvious on review

apps/core/lib/core/ecto/date_range.ex
kiurchv: In my opinion, braces are good to distinguish built-in/well-known type names from others

PR #4427, opened 2019-03-10
refactor(*): return timestamps with timezones

kiurchv:


apps/core/lib/core/black_list_users/black_list_user.ex
```diff
@@ -12,6 +12,6 @@ defmodule Core.BlackListUsers.BlackListUser do
 
     has_many(:parties, Core.Parties.Party, foreign_key: :tax_id, references: :tax_id)
 
-    timestamps()
+    timestamps(type: :utc_datetime)
```
AlexKovalevych: Why do you need that and  what about existing data?

AlexKovalevych: Also, would be nice to have a link to appropriate ticket in the PR description (and yes, i can find it from the branch name)

PR #4462, opened 2019-03-14
feat(graphql): implement create immnDosage

progsmile:
#4185


apps/core/lib/core/medications/medications.ex
```diff
@@ -396,14 +396,14 @@ defmodule Core.Medications do
 
   # Create
 
-  def create_innm_dosage(attrs, headers), do: create_medication_entity(INNMDosage, attrs, headers)
+  def create_innm_dosage(attrs, actor_id), do: create_medication_entity(INNMDosage, attrs, actor_id)
```
PavelVesnin: for consistency, actor_id should be a first argument

progsmile: suppose it's unlogical to put actor_id first

progsmile: in elixir data always first, than some config stuff

PavelVesnin: than it should be everywhere. But now actor_id is first argument

progsmile: as you may notice there are more cases with `actor_id` last in `Medications`, so no consistency there

PavelVesnin: https://github.com/edenlabllc/ehealth.api/blob/develop/apps/core/lib/core/medical_programs/medical_programs.ex#L47
https://github.com/edenlabllc/ehealth.api/blob/develop/apps/core/lib/core/medications/medications.ex#L551
and others

Issue #4549, opened 2019-03-25
Allow msp_pharmacy create contract request


AlekzandrV:
As a msp_pharmacy owner
I want be able to create both capitation and reimbursement contract
So that i can manage the amount of service and compensation separately

- [ ] update create contract request - allow legal entity with type `MSP_PHARMACY` to create contract request (capitation, reimbursement)
- [ ] improve division validation - validate division type referenced to contract request depending on contract type

contract type | division type
----------------|----------------------
capitation | ["CLINIC", "AMBULANT_CLINIC", "FAP"]
reimbursement | ["DRUGSTORE", "DRUGSTORE_POINT"]

PR #4563, opened 2019-03-26
feat(ehealth): added MSP_PHARMACY for ContractRequest creation

PavelVesnin:
#4549
- added MSP_PHARMACY in allowed legal entities types for contract_request creation
- added division type validation for contract requests


apps/core/config/config.exs
```diff
@@ -264,6 +264,10 @@ config :core, :legal_entity_division_types,
     {:system, :list, "LEGAL_ENTITY_MSP_PHARMACY_DIVISION_TYPES",
      ["CLINIC", "AMBULANT_CLINIC", "FAP", "DRUGSTORE", "DRUGSTORE_POINT"]}
 
+config :core, :contracts_division_types,
+  capitation: {:system, :list, "CAPITATION_CONTRACT_DIVISION_TYPES", ["CLINIC", "AMBULANT_CLINIC", "FAP"]},
```
progsmile: Let's remove default values?

AlexKovalevych: why?

PavelVesnin: it's base logic for contracts. Values can be configured, but probably wouldn't

apps/ehealth/test/web/controllers/contract_request/capitation_controller_test.exs
```diff
@@ -1430,6 +1430,135 @@ defmodule EHealth.Web.ContractRequest.CapitationControllerTest do
     end
   end
 
+  describe "create contract request by MSP_PHARMACY" do
+    setup %{conn: conn} do
+      msp_pharmacy()
+
+      stub(MediaStorageMock, :create_signed_url, fn _, _, resource, _, _ ->
```
progsmile: Can we use `expect`? And in cases below?

docs/ENVIRONMENT.md
```diff
@@ -164,3 +164,10 @@ Also sample `.env` can be used as payload for `docker run` cli.
 | -------------------------------------------- | ------------- | ----------- |
 | LEGAL_ENTITY_MSP_EMPLOYEE_TYPES              | not set       | List of available Employee types in Legal Entity
 | LEGAL_ENTITY_PHARMACY_EMPLOYEE_TYPES         | not set       | List of available Pharmacy Employee types in Legal Entity
+ored
+
+# Legal Entity division types for contracts
+| VAR_NAME                                     | Default Value                      | Description |
+| -------------------------------------------- | ---------------------------------- | ----------- |
+| CAPITATION_CONTRACT_DIVISION_TYPES           | "CLINIC", "AMBULANT_CLINIC", "FAP" | List of allowed Division types for Capitation Contract Request creation
```
progsmile: Table is not closed

apps/core/lib/core/contract_requests/validator.ex
```diff
@@ -391,17 +400,17 @@ defmodule Core.ContractRequests.Validator do
     end
   end
 
-  def validate_contractor_divisions(%{
+  def validate_contractor_divisions(type, %{
```
progsmile:
```suggestion
  def validate_contractor_divisions(contract_request_type, %{
```

apps/core/config/config.exs
progsmile: So application wouldn't start without env vars 
@PavelVesnin  Do you need to substitute values? We may use it without env var

AlexKovalevych: you want it to not start ? why? it's a good practice to allow configure values which may be changed in the future

apps/ehealth/test/web/controllers/contract_request/capitation_controller_test.exs
PavelVesnin: No, because tests with invalid params not used all mocks and for DRY I put it in setup.
Case with mocks already tested multiple times

apps/core/config/config.exs
PavelVesnin: No, values should be configurable

progsmile: @AlexKovalevych because we can easily miss env var and use default params thinking that everything is ok

progsmile: They may be default in test, dev envs for convenience, but not prod

PR #4583, opened 2019-03-26
feat(*): refactor Employee deactivation logic, add `deactivateEmployee` mutation

kiurchv:


apps/core/lib/core/api/mithril.ex
```diff
@@ -123,8 +123,8 @@ defmodule Core.API.Mithril do
     delete!("/admin/users/#{user_id}/roles/#{role_id}", headers)
   end
 
-  def delete_user_roles_by_user_and_role_name(user_id, role_name, headers) do
-    delete!("/admin/users/#{user_id}/roles?role_name=#{role_name}", headers)
```
PavelVesnin: headers must be there, because it contains request_id that used for logging cross-services requests

apps/core/lib/core/employees/employee_updater.ex
```diff
@@ -23,45 +24,37 @@ defmodule Core.Employees.EmployeeUpdater do
 
   @producer Application.get_env(:core, :kafka)[:producer]
 
-  def deactivate(%{"id" => id} = params, headers, with_owner \\ false) do
-    deactivate(id, params["legal_entity_id"], "auto_employee_deactivate", headers, with_owner)
-  end
+  @message_does_not_belongs "Employees not belonging to the current legal entity can't be deactivated"
```
BatrakM: does_not_belong

apps/jobs/test/unit/merge_legal_entity_job_test.exs
```diff
@@ -98,9 +98,9 @@ defmodule Unit.LegalEntityMergeJobTest do
       assert @dismissed == Employees.get_by_id(employee_dismissed2.id).status
       assert @dismissed == Employees.get_by_id(employee_dismissed3.id).status
 
-      assert "auto_reorganization" == Employees.get_by_id(employee_dismissed.id).status_reason
-      assert "auto_reorganization" == Employees.get_by_id(employee_dismissed2.id).status_reason
-      assert "auto_reorganization" == Employees.get_by_id(employee_dismissed3.id).status_reason
+      assert "auto_legal_entity_reorganization" == Employees.get_by_id(employee_dismissed.id).status_reason
+      assert "auto_legal_entity_reorganization" == Employees.get_by_id(employee_dismissed2.id).status_reason
+      assert "auto_legal_entity_reorganization" == Employees.get_by_id(employee_dismissed3.id).status_reason
```
PavelVesnin: why not uppercase?

apps/core/lib/core/employees/employee_updater.ex
```diff
@@ -23,45 +24,37 @@ defmodule Core.Employees.EmployeeUpdater do
...
+         :ok <- maybe_deactivate_declarations(employee, reason, actor_id) do
+      set_employee_status_as_dismissed(employee, reason, actor_id, skip_contracts_suspend?)
     end
   end
 
-  def check_transition(%Employee{employee_type: @type_owner}, false) do
-    {:error, {:conflict, "Owner can’t be deactivated"}}
-  end
+  defp check_belongingness(%Employee{legal_entity_id: client_id}, client_id), do: :ok
+  defp check_belongingness(_, _), do: {:error, {:forbidden, @message_does_not_belongs}}
```
BatrakM: Sound gross. Maybe just defp belongs?(...

apps/jobs/test/unit/merge_legal_entity_job_test.exs
kiurchv: Agree, good idea

apps/core/lib/core/api/mithril.ex
kiurchv: Hmm, but jobs don't pass any headers here: https://github.com/edenlabllc/ehealth.api/blob/19079c66395546ec051d441144dc0a059480d84e/apps/jobs/lib/jobs/legal_entity_deactivation_job.ex#L75

apps/core/lib/core/employees/employee_updater.ex
kiurchv: I'd expect to receive boolean from function named `belongs?`

BatrakM: Fair enough )

apps/core/lib/core/api/mithril.ex
PavelVesnin: ok, than pass from job empty list.
Headers for API calls is mandatory

apps/jobs/lib/jobs/legal_entity_deactivation_job.ex
```diff
@@ -26,7 +26,7 @@ defmodule Jobs.LegalEntityDeactivationJob do
   @status_active LegalEntity.status(:active)
   @status_closed LegalEntity.status(:closed)
 
-  @status_reason "AUTO_DEACTIVATION_LEGAL_ENTITY"
+  @status_reason "AUTO_LEGAL_ENTITY_DEACTIVATION"
```
PavelVesnin: do we need update status_reason on prod?
let's check it in DB

apps/core/lib/core/employees/employee_updater.ex
```diff
@@ -23,45 +24,37 @@ defmodule Core.Employees.EmployeeUpdater do
...
 
-  def do_deactivate(employee, reason, headers, user_id, skip_contracts_suspend?) do
-    deactivation_event = %{"employee_id" => employee.id, "actor_id" => user_id, "reason" => reason}
-
+  def deactivate(employee, reason, headers, actor_id, skip_contracts_suspend?) do
     with active_employees <- get_active_employees(employee),
-         :ok <- revoke_user_auth_data(employee, active_employees, headers),
-         :ok <- @producer.publish_deactivate_declaration_event(deactivation_event) do
-      set_employee_status_as_dismissed(employee, reason, user_id, skip_contracts_suspend?)
+         :ok <- maybe_revoke_user_auth_data(employee, active_employees, headers),
```
progsmile: suppose that's bad naming with `maybe_` prefix
Let's remove that!

PavelVesnin: you're not first time mentioned, that it's bad naming. Why?
It's clearly to understand, that function has optional logic

progsmile: There are so many optional logic: each `match`, `if` adds some optionality - that doesn't mean we should add `maybe_` everywhere. Such name make it unconfident

PR #4598, opened 2019-03-27
feat: push event to event manager via kafka

marinakr:


apps/graphql/features/contexts/context.exs
```diff
@@ -2160,20 +2157,6 @@ defmodule GraphQL.Features.Context do
     end
   )
 
-  then_(
-    ~r/^event manager has event for (?<entity_type>[^"]+) with ID "(?<entity_id>[^"]+)" and consumer ID "(?<updated_by>[^"]+)"$/,
```
progsmile: BDD tests would pass
Find and remove lines that contains `event manager has event for` in *.feature files

marinakr: `cat apps/graphql/features/*/*.feature | grep event`
Output:
   ``` And event manager has event for CapitationContract with ID <database_id> and consumer ID <consumer_id>```
    ```And event manager has event for ReimbursementContract with ID <database_id> and consumer ID <consumer_id>```

progsmile: Remove that lines

apps/core/lib/core/contracts/contracts.ex
```diff
@@ -80,7 +80,7 @@ defmodule Core.Contracts do
                contract
                |> changeset(%{"status" => @status_terminated})
                |> PRMRepo.update() do
-          EventManager.insert_change_status(contract, @status_terminated, user_id)
+          EventManager.puplish_change_status(contract, @status_terminated, user_id)
```
progsmile: Use `publish` instead of `puplish`

apps/graphql/features/contexts/context.exs
```diff
@@ -5,7 +5,7 @@ defmodule GraphQL.Features.Context do
   import Core.Expectations.Mithril, only: [get_client_type_name: 1]
   import Core.Factories, only: [build: 2, build_list: 2, insert: 3, insert_list: 3]
   import Ecto.Query, only: [where: 2]
-  import Mox, only: [expect: 3, expect: 4]
+  import Mox, only: [expect: 3, expect: 4, set_mox_global: 0]
```
progsmile: No need of `set_mox_global: 0`

apps/graphql/features/contexts/context.exs
```diff
@@ -2029,6 +2029,51 @@ defmodule GraphQL.Features.Context do
     end
   )
 
+  when_(
+    ~r/^I suspend (?<contract_type>\w+) contract and event where databaseId is "(?<database_id>[^"]+)"$/,
+    fn %{conn: conn}, %{database_id: database_id, contract_type: contract_type} ->
+      expect(KafkaMock, :publish_to_event_manager, fn _ -> :ok end)
```
progsmile: We will reuse this expect, so let's move it to other particular step, like:
`And event should be published to event manager`

mix.exs
```diff
@@ -31,4 +32,10 @@ defmodule Ehealth.MixProject do
       {:git_ops, git: "https://github.com/AlexKovalevych/git_ops.git", branch: "fix_version_compare", only: [:dev]}
     ]
   end
+
+  defp aliases do
+    [
+      test: ["test", fn _ -> Mix.shell().cmd("cd apps/graphql && mix white_bread.run && cd ../../") end]
```
progsmile: That would not work with error on CI, please return as it was

marinakr: it successfully works

apps/graphql/features/contexts/context.exs
```diff
@@ -2032,6 +2029,51 @@ defmodule GraphQL.Features.Context do
     end
   )
 
+  when_(
+    ~r/^I suspend (?<contract_type>\w+) contract where databaseId is "(?<database_id>[^"]+)" and event published$/,
```
progsmile: No. We don't have to write `and event published` each time, it's atomic operation and should be defined separately

mix.exs
progsmile: @PavelVesnin need your thoughts on this

apps/graphql/features/contexts/context.exs
marinakr: actually  it published only in to success cases, and has separate match

progsmile: this match is big and this case with event manager is often used

marinakr: 2 times is not often. It has correct name and correct match

apps/core/lib/core/kafka/producer.ex
```diff
@@ -3,10 +3,19 @@ defmodule Core.Kafka.Producer do
...
   @edr_verification_events_topic "edr_verification_events"
+  @event_manager_topic "event_manager_topic"
+
   @behaviour Core.Kafka.ProducerBehaviour
 
   require Logger
 
+  def publish_to_event_manager(event) do
+    with :ok <- Kaffe.Producer.produce_sync(@event_manager_topic, 0, "", :erlang.term_to_binary(event)) do
+      Logger.info("Published event #{inspect(event)} to kafka")
```
AlexKovalevych: probably we shouldnt log all events, just fails

Issue #4669, opened 2019-04-03
Prod error 500 bounce from [email] not found


lymychp:
```  "app": {
      "log": "[PostmarkClient]: bounce from '[redacted]' not found",
      "request_id": "8461a9fb-994a-43ea-9354-c9d615e64bb5#1280141",
      "severity": "ERROR",
      "sourceLocation": {
        "file": "lib/core/email/postmark.ex",
        "line": 19,
        "function": "Elixir.Core.Email.Postmark.activate_email/1"
      }
    },
    "request_id": "8461a9fb-994a-43ea-9354-c9d615e64bb5#1280141",
    "tag": "il-api",
    "@timestamp": "2019-04-03T07:10:41.384042708+00:00"
```

progsmile:
Log severity has been changed to warn

PR #4717, opened 2019-04-09
chore(core): fix log severity on postmark calls

progsmile:
#4669


apps/core/lib/core/email/postmark.ex
```diff
@@ -15,12 +15,8 @@ defmodule Core.Email.Postmark do
 
       {:ok, bounce_id}
     else
-      {:error, error} = error_result ->
-        Logger.error("[PostmarkClient]: #{error}")
-        error_result
-
       error ->
-        Logger.error("[PostmarkClient]: error happened #{inspect(error)}")
+        Logger.info("[PostmarkClient]: #{inspect(error)}")
```
BatrakM: So now you will have {:error, error} tuple in logs? Or {:error, error} path never matched?

progsmile: Yes, I've decided not to match error result, only for log message separation. Do you think it's worth to do?

BatrakM: Well, that's not a lot of code to keep for a more clean error log, but on the other side 10 extra symbols in log message is not a bit deal. Let's hear some other opinions maybe

AlexKovalevych: why not warn?

progsmile: yep, warn is better

progsmile: Because we can't do nothing regarding that

progsmile: Agree?

PR #4768, opened 2019-04-10
fix(*): use the same format for validation errors

progsmile:


apps/core/lib/core/validators/json_objects.ex
```diff
@@ -63,11 +63,13 @@ defmodule Core.Validators.JsonObjects do
     if Enum.any?(keys, &(&1 == required_item)) do
       :ok
     else
-      %ValidationError{
-        description: "Must contain required item.",
-        path: get_path(object_path, "[]", key_name),
-        params: [required_item]
-      }
+      Enum.map(0..(length(keys) - 1), fn index ->
```
progsmile: `array_item_required` now returns list of errors as it's done in json schema v6 and anyOf rule

PR #4771, opened 2019-04-10
feat: mpi search  rpc only, #6019

marinakr:


marinakr:
Merge with https://github.com/edenlabllc/mpi.api/pull/294

PR #4786, opened 2019-04-12
chore: add party documents fields #6131

marinakr:


apps/core/lib/core/parties/document.ex
```diff
@@ -4,21 +4,21 @@ defmodule Core.Parties.Document do
...
-  )a
+  @fields_optional ~w(issued_at issued_by)a
+  @fields_required ~w(type number)a
 
   @derive {Jason.Encoder, except: [:__meta__]}
   @primary_key false
   schema "documents" do
     field(:type, :string)
     field(:number, :string)
+    field(:issued_at, :string)
```
BatrakM: in employee_request_create_request.json this field is described like this

"issued_at": {
    "type": "string",
    "format": "date"
}

so maybe it can be stored as date

PR #4799, opened 2019-04-15
fix: create contract request fixes

kiurchv:


apps/graphql/features/contexts/context.exs
```diff
@@ -2754,6 +2757,8 @@ defmodule GraphQL.Features.Context do
     |> String.to_atom()
   end
 
+  defp prepare_value(_, _, "null"), do: {:ok, nil}
```
PavelVesnin: what?

kiurchv: we should ignore introspection results when preparing `null` values

Issue #4475, opened 2019-03-18
create contract request by NHS


svetavedmed:
- [conf](https://edenlab.atlassian.net/wiki/x/PIUVKw)
- graphQL - create contract request mutation

svetavedmed:
- [ ] on create contract error  “message”: “Employee division must be unique”, we should use Employee w/o end_date olny.
- [ ] after contract request create there is no `assignee`

svetavedmed:
- [ ] empty NHS fields after create contract request
```
issueCity: null
miscellaneous: null
nhsContractPrice: null
nhsLegalEntity: null
nhsPaymentMethod: null
nhsSigner: null
nhsSignerBase: null
previousRequest: null
```

svetavedmed:
- [ ] no documents in any type of contracts. Like there are no links for contract request even create_from_draft
- [ ] null in BD contract_request.data for new contract_request cretae_from_contract. It is created in status `Approved` and should be created with data

PR #4928, opened 2019-05-07
fix(graphql): copy contract request files on create from contract process

progsmile:
#4475


apps/core/lib/core/contract_requests/contract_requests.ex
```diff
@@ -200,6 +200,8 @@ defmodule Core.ContractRequests do
          %Changeset{valid?: true} = changes <- changeset(contract_request, insert_params),
          data <- render_contract_request_data(changes),
          %Changeset{valid?: true} = changes <- put_change(changes, :data, data),
+         :ok <- save_signed_content(id, params, headers, "signed_content/contract_request_created_from_contract"),
```
PavelVesnin: is it required statement for bucket name `signed_content/contract_request_created_from_contract` ?

progsmile: @kiurchv, answer the question please.

PavelVesnin: I think that it should be the same name as for contract_request. 
I see no reason for new bucket for copied contract request. It's just an another way to create contract request

apps/core/lib/core/contract_requests/contract_requests.ex
```diff
@@ -186,8 +187,7 @@ defmodule Core.ContractRequests do
          :ok <- validate_contractor_divisions_dls(pack.type, content["contractor_divisions"] || []),
          :ok <- validate_start_date_year(content),
          :ok <- validate_end_date(content),
-         :ok <- create_validate_contractor_owner_id(pack.type, content),
-         :ok <- save_signed_content(id, params, headers, "signed_content/contract_request_created_from_contract"),
+         :ok <- validate_contractor_owner_id_on_create(pack.type, content),
```
PavelVesnin:
```suggestion
         :ok <- validate_contractor_owner_id(pack.type, content),
```

progsmile: such function already exists

apps/core/lib/core/contract_requests/contract_requests.ex
progsmile: yep, agree

kiurchv: Since signed content of the contract request created from the draft and the one created from the parent contract are different I propose to use different buckets for them

Issue #4909, opened 2019-05-02
Update Create related Legal Entities


mamaelka:
- [ ] Validate merged_from LE status 
- [ ] Validate merged_from <> merged_to
- [ ] Validate LE type transitions
- [ ] Update merged_from legal entity status to 'Reorganised'
- [ ] Update scopes for reorganised LE, set same scopes as for MSP_Limited client type
[Confluence] (https://edenlab.atlassian.net/wiki/spaces/EH/pages/635568333/Create+related+Legal+Entities)

PR #4946, opened 2019-05-08
feat: merge LE changes

kiurchv:
connects #4909


apps/core/test/support/prm_factories/legal_entity_factory.ex
```diff
@@ -52,14 +52,18 @@ defmodule Core.PRMFactories.LegalEntityFactory do
         }
       end
 
-      def related_legal_entity_factory do
-        %RelatedLegalEntity{
+      def related_legal_entity_factory(attrs) do
+        record = %RelatedLegalEntity{
```
PavelVesnin: I think that `entity` is more preferable than `record`

kiurchv: We use `record` since ExMachina uses this term itself

apps/graphql/features/contexts/context.exs
```diff
@@ -234,7 +234,22 @@ defmodule GraphQL.Features.Context do
     fn %{content: content} = state, %{table_data: table_data} ->
       signers = Enum.map(table_data, &prepare_input_attrs/1)
 
-      expect_signed_content(content, signers)
+      stub(SignatureMock, :decode_and_validate, fn _, _, _ ->
```
progsmile: How about using `expect` depending on something (maybe `signers` count)? 
Or there is no way to determine mock execution?

kiurchv: If the request fails in the middleware this expect will be never called

Issue #4464, opened 2019-03-14
migrate verification,ael,man,ds calls to rpc


AlexKovalevych:

progsmile:
Everything is in pulls

PR #4952, opened 2019-05-10
feat: replace rest calls with rpc on man and otp verification services

progsmile:
#4464


apps/core/lib/core/declaration_requests/api/v1/creator.ex
```diff
@@ -798,8 +798,8 @@ defmodule Core.DeclarationRequests.API.V1.Creator do
       {:ok, printout_content} ->
         put_change(changeset, :printout_content, printout_content)
 
-      {:error, error_response} ->
-        add_error(changeset, :printout_content, format_error_response("MAN", error_response))
+      {:error, _} ->
+        add_error(changeset, :printout_content, format_error_response("MAN", "Remote server internal error"))
```
imax-iva: may need to raise Exception?

apps/core/lib/core/man/templates/contract_request_printout_form.ex
```diff
@@ -29,10 +29,9 @@ defmodule Core.Man.Templates.CapitationContractRequestPrintoutForm do
 
   def render(
         %CapitationContractRequest{parent_contract_id: parent_contract_id, contract_number: contract_number} =
-          contract_request,
-        headers
+          contract_request
       )
-      when not is_nil(parent_contract_id) do
+      when parent_contract_id != nil do
```
imax-iva: why? :)

apps/core/lib/core/declaration_requests/api/v1/creator.ex
progsmile: support old behaviour

apps/core/lib/core/man/templates/contract_request_printout_form.ex
progsmile: suppose formatter format in single line

PR #4990, opened 2019-05-15
feat: add `updateLegalEntityStatus` mutation

kiurchv:


apps/core/lib/core/legal_entities/legal_entity.ex
```diff
@@ -63,6 +63,7 @@ defmodule Core.LegalEntities.LegalEntity do
     field(:public_name, :string)
     field(:short_name, :string)
     field(:status, :string)
+    field(:status_reason, :string)
```
PavelVesnin: where is `reason` field?

apps/core/priv/prm_repo/migrations/20190513141500_add_status_update_attrs_to_legal_entities.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Core.PRMRepo.Migrations.AddStatusUpdateAttrsToLegalEntities do
+  use Ecto.Migration
+
+  def change do
+    alter table(:legal_entities) do
+      add(:reason, :text)
+      add(:status_reason, :string)
```
PavelVesnin: why not `:text`?

PR #4992, opened 2019-05-15
feat: replace rest call with rpc in digital signature

progsmile:
#4464


apps/core/lib/core/api/signature.ex
```diff
@@ -107,6 +119,22 @@ defmodule Core.API.Signature do
     ResponseDecoder.check_response(%HTTPoison.Response{body: data, status_code: 422})
   end
 
+  defp wrap_rpc_response(content, signatures) do
```
AlexKovalevych: i don't think we need to wrap rpc response to rest-like response, why not match rpc response directly?

progsmile: Module behaviour should remain the same with enabled / disabled signature. Now if it's disabled - rest response is returned. Look at `SignatureValidator.validate` usage, in case of mismatch it return response to client (in most cases). So I decided to replace http call only, not to break anything.

AlexKovalevych: > Module behaviour should remain the same with enabled / disabled signature.

correct, but now it's not a rest response, it's rpc response and you should handle that change in all places which use ds response

progsmile: In error case rest response is send to client, there are many places to handle. Suppose it's another task

PR #5016, opened 2019-05-20
test(graphql): add features for program services, services and service groups

novohatnya:


kiurchv:
Was split into separate PRs

Issue #4833, opened 2019-04-18
sign MRR - convert not a base64 string to 4xx error


lymychp:
1. logs from prod
```
"app": {
      "log": "[{%{description: \"Not a base64 string\", params: [], rule: :invalid}, \"$.signed_content\"}]",
      "request_id": "ec5eb096-828d-41e8-9618-572fa8f77e27#34205",
      "severity": "ERROR",
      "sourceLocation": {
        "file": "lib/api/web/controllers/digital_signatures_controller.ex",
        "line": 26,
        "function": "Elixir.API.Web.APIController.index/2"
      }
    },
    "request_id": "ec5eb096-828d-41e8-9618-572fa8f77e27#34205",
    "tag": "ds-api",
    "@timestamp": "2019-04-18T05:25:17.470639128+00:00"
  },
```

2. another example
```
 "app": {
      "log": ":function_clause, [{Core.MedicationRequestRequest.OperationHelpers, :add_changeset_error, [{:error, %{\"error\" => %{\"invalid\" => [%{\"entry_type\" => \"content\", \"rules\" => [%{\"params\" => %{\"content\" => \"<<48, 130, 3, 196, 2, 1, 3, 48, 130, 3, 97, 6, 9, 42, 134, 72, 134, 247, 13, 1, 7, 1, 160, 130, 3, 82, 4, 130, 3, 78, 48, 130, 3, 74, 48, 130, 1, 149, 6, 9, 42, 134, 72, 134, 247, 13, 1, 7, 1, 160, ...>>\"}, \"rule\" => \"json\"}]}], \"message\" => \"Malformed encoded content. Probably, you have encoded corrupted JSON. Error: %Jason.DecodeError{data: <<48, 130, 3, 196, 2, 1, 3, 48, 130, 3, 97, 6, 9, 42, 134, 72, 134, 247, 13, 1, 7, 1, 160, 130, 3, 82, 4, 130, 3,  (truncated)",
      "request_id": "20e57a1d-8af2-4765-9e7d-029b558ee75f#31556",
      "severity": "INFO",
      "sourceLocation": {
        "file": "lib/instruments.ex",
        "line": 10,
        "function": "Elixir.LoggerJSON.Phoenix.Instruments.phoenix_error_render/3"
      }
    },
    "request_id": "20e57a1d-8af2-4765-9e7d-029b558ee75f#31556",
    "tag": "il-api",
[...]

PR #5049, opened 2019-05-23
fix: convert not a base64 string to 4xx error on sign MRR

imax-iva:
https://github.com/edenlabllc/ehealth.api/issues/4833


apps/core/lib/core/medication_request_requests/operation_helpers.ex
```diff
@@ -259,14 +259,30 @@ defmodule Core.MedicationRequestRequest.OperationHelpers do
...
       :signed_medication_request_request,
       reason,
       [validation: :invalid]
     ])
   end
 
+  defp add_changeset_error({:invalid_signature, reason}, operation) do
+    Operation.call_changeset(operation, &add_error/4, [
+      :signed_medication_request_request,
+      to_string(reason),
```
progsmile: how about `inspect(reason)` ?

PR #5061, opened 2019-05-24
feat(core): created table `program_services` in PRM

PavelVesnin:


apps/core/priv/prm_repo/migrations/20190524102703_create_program_services_table.exs
```diff
@@ -0,0 +1,20 @@
+defmodule Core.PRMRepo.Migrations.CreateProgramServicesTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:program_services, primary_key: false) do
+      add(:id, :uuid, primary_key: true)
+      add(:program_id, :uuid, null: false)
```
kiurchv: We should add foreign key constraints here
```suggestion
      add(:program_id, references(:medical_programs,  type: :uuid), null: false)
```

apps/core/priv/prm_repo/migrations/20190524102703_create_program_services_table.exs
```diff
@@ -0,0 +1,20 @@
...
+      add(:service_id, :uuid)
```
kiurchv:
```suggestion
      add(:service_id, references(:services,  type: :uuid))
```

apps/core/priv/prm_repo/migrations/20190524102703_create_program_services_table.exs
```diff
@@ -0,0 +1,20 @@
...
+      add(:service_group_id, :uuid)
```
kiurchv:
```suggestion
      add(:service_group_id, references(:service_groups,  type: :uuid))
```

Issue #4899, opened 2019-04-26
Return non 500 response for invalid content


dhlebnikov1:
request_id to find logs - `9233d254-9efb-40a3-9fed-709ae4a3b79b#1351394`
when client send invalid content we process it and give answer with
```
{
  "invalid": [
    {
      "entry_type": "request",
      "rules": [
        {
          "rule": "json"
        }
      ]
    }
  ],
  "message": "Malformed request. Probably, you have sent corrupted JSON.",
  "type": "request_malformed"
}
```
but `il` return to client error `500` code 
```
:function_clause, [{Core.MedicationRequestRequest.OperationHelpers, :add_changeset_error, [{:error, %{"error" => %{"invalid" => [%{"entry_type" => "content", "rules" => [%{"params" => %{"content" => "<<48, 130,...
```

PR #5147, opened 2019-05-31
fix: correct handling ds.api errors (base64, json format etc)

imax-iva:
https://github.com/edenlabllc/ehealth.api/issues/4899


apps/core/lib/core/declaration_requests/api/v1/sign.ex
```diff
@@ -87,6 +87,8 @@ defmodule Core.DeclarationRequests.API.Sign do
     |> add_error(:signed_declaration_request, "#{code}: #{message}")
   end
 
+  def normalize_signature_error({:error, _} = error), do: error
```
AlexKovalevych: not needed

PR #5226, opened 2019-06-10
Terminate employee and person endpoint via kafka#4242

imax-iva:


apps/jobs/lib/jobs/declaration_termination_job.ex
```diff
@@ -0,0 +1,130 @@
...
+  def get_job(id) do
+    case JabbaClient.get_job(id) do
+      {:ok, job} -> {:ok, job}
+      nil -> {:ok, nil}
+    end
+  end
+
+  def create(processed_data) do
+    tasks = get_terminate_declaration_tasks(processed_data)
+    JabbaClient.create_job(tasks, @declaration_termination_job_type, name: "Terminate declaration")
```
PavelVesnin:
```suggestion
    opts = [
      name: "Terminate declaration",
      callback: {"ehealth", Jobs.DeclarationTerminationJob, :job_processed, []},
      strategy: "CONCURRENT"
    ]  
      JabbaClient .create_job(tasks, @declaration_termination_job_type, opts)
```
Callback will called when job processed or failed.
So you don't need indexes in tasks anymore.

